PR #9, opened 2022-11-27
'Solution'

Sebshe:


app/main.py
```diff
@@ -1,25 +1,103 @@
...
 from abc import ABC
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Slide, name: str) -> None:
```
Serhii-Leonenko, line 11:
```suggestion
    def __set_name__(self, owner: object, name: str) -> None:
```
It's better to use object, for example you can validate another class's data, it's a universal class. The same in other methods.

app/main.py
```diff
@@ -1,25 +1,103 @@
...
+            )
+        except ValueError:
+            return False
+        return True
+
+
+baby = Visitor(name="t", age=10, weight=30, height=120)
+slide = Slide(name="baby", limitation_class=ChildrenSlideLimitationValidator)
+
+print(slide.can_access(baby))
```
Serhii-Leonenko, line 103: Don't push data, which you use for local testing

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,25 +1,97 @@
...
+    height = IntegerRange(min_amount=80, max_amount=120)
+    weight = IntegerRange(min_amount=20, max_amount=50)
+
+    def __init__(
+            self,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        super().__init__(age, weight, height)
```
Grayder0152, line 62: it is not necessary block of code

app/main.py
```diff
@@ -1,25 +1,97 @@
...
+    height = IntegerRange(min_amount=120, max_amount=220)
+    weight = IntegerRange(min_amount=50, max_amount=120)
+
+    def __init__(
+            self,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        super().__init__(age, weight, height)
```
Grayder0152, line 76: it is not necessary block of code

Grayder0152 (approved):

PR #15, opened 2022-12-07
Solution

aarrtemm:


app/main.py
```diff
@@ -1,25 +1,69 @@
...
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: object) -> str:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, True)
+        else:
```
Abnormaltype, line 19: unnecessary else

Abnormaltype (changes requested):

app/main.py
aarrtemm: Please tell me the best way to do it. Because I've already tried everything, I fail the tests

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
```
Abnormaltype, line 36:
```suggestion
class SlideLimitationValidator(ABC):
    def __init__(self, age: int, height: int, weight: int) -> None:
        self.age = age
        self.height = height
        self.weight = weight
```

app/main.py
```diff
@@ -1,25 +1,71 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
     pass
```
Abnormaltype, line 40: unnecessary

Abnormaltype (changes requested):

app/main.py
aarrtemm: I don't quite understand what needs to be removed.

Abnormaltype: pass

OleksandrPozden (approved):
Nice!

PR #19, opened 2022-12-12
Solution

6aPa6aH:


app/main.py
```diff
@@ -2,24 +2,66 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.public_name = name
```
IvanRamyk, line 10: it seems that you don't use public name

why do you need it?

IvanRamyk (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #34, opened 2023-01-01
Solution

Tania-Kharchuk:


app/main.py
```diff
@@ -1,25 +1,86 @@
...
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+        self,
+        name: str,
+        limitation_class: callable
```
olena-hul, line 76: I think it can be annotated with `SlideLimitationValidator` instead

olena-hul (changes requested):

app/main.py
Tania-Kharchuk: Do you mean SlideLimitationValidator instead callable?

olena-hul: Yes

Abnormaltype (approved):
Well done!

PR #36, opened 2023-01-03
solution implemented

drapail:


app/main.py
```diff
@@ -2,24 +2,68 @@
 
 
 class IntegerRange:
-    pass
+
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: object, owner: object) -> object:
```
olena-hul, line 10: Use real types for annotations, not `object`

app/main.py
```diff
@@ -2,24 +2,68 @@
...
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            self.protected_name = value
+            return
+        raise ValueError
```
olena-hul, line 17: I think you can change condition to opposite and raise `ValueError` inside `if`. And outside `if` you can set the value for `protected_name`. This way we can avoid redundant return

app/main.py
```diff
@@ -2,24 +2,68 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
olena-hul, line 45: This `__init__` is redundant. The implementation will be inherited from base class

app/main.py
```diff
@@ -2,24 +2,68 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
olena-hul, line 54: The same comment about `__init__` applies here

app/main.py
```diff
@@ -2,24 +2,68 @@
...
 
 
 class Slide:
-    pass
+
+    def __init__(self, name: str, limitation_class: SlideLimitationValidator) \
+            -> None:
```
olena-hul, line 60: If the method definition doesn't fit into 1 line, place each argument in separate lines and avoid using `\`

olena-hul (changes requested):

app/main.py
drapail: Great comment, this way code looks much nicer, thanks!

olena-hul (approved):

PR #40, opened 2023-01-06
'Solution'

artur-leleiko:


app/main.py
```diff
@@ -2,24 +2,61 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        validator = self.limitation_class(
+            visitor.age,
+            visitor.weight,
+            visitor.height
+        )
+        return len(validator.__dict__) == 3
```
YuriiKindrat, line 62: it is better to raise exception in descriptor __set__ method and then catch it here -> return True/False

YuriiKindrat (changes requested):

andrewbelo (approved):
gj

PR #50, opened 2023-01-20
'Solution'

Sergey-Bushinskiy:


app/main.py
```diff
@@ -1,25 +1,98 @@
...
+
+    def __set_name__(self, owner: str, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: str, owner: str) -> None:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: str, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise \
```
YuriiKindrat, line 18: unnecessary " \ " here, try not use it

app/main.py
```diff
@@ -1,25 +1,98 @@
...
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(
+            self,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        super().__init__(age, weight, height)
```
YuriiKindrat, line 63: no need to rewrite init method here

app/main.py
```diff
@@ -1,25 +1,98 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(
+            self,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        super().__init__(age, weight, height)
```
YuriiKindrat, line 77: the same

PR #51, opened 2023-01-20
'Solution'

Bloodviel:


app/main.py
```diff
@@ -1,25 +1,68 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: Union[ChildrenSlideLimitationValidator,
+                                         AdultSlideLimitationValidator]
+                 ) -> None:
```
Vasyl-Smutok, line 59: better like this
```
def __init__(
        self,
        name: str,
        limitation_class: Union[
            ChildrenSlideLimitationValidator,
            AdultSlideLimitationValidator
        ]
) -> None:
```

app/main.py
```diff
@@ -1,25 +1,68 @@
...
 
 
 class Visitor:
-   pass
+    def __init__(self,
+                 name: str,
+                 age: int,
+                 weight: int,
+                 height: int
+                 ) -> None:
```
Vasyl-Smutok, line 29: you cat write like this
`    def __init__(self, name: str, age: int, weight: int, height: int) -> None:`
or 
```
    def __init__(
            self, 
            name: str, 
            age: int, 
            weight: int, 
            height: int
    ) -> None:
```

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,25 +1,73 @@
...
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: Union[
+                ChildrenSlideLimitationValidator,
+                AdultSlideLimitationValidator
+            ]
```
ArsenPidhoretskyi, line 63: When you add an additional Validator, you must change typing. I recommend you to set the type as:
Btw, it is not the correct type, because you pass a class, not an instance

app/main.py
```diff
@@ -1,25 +1,73 @@
...
+
+    def __set_name__(self, owner: IntegerRange, name: str) -> None:
+        self._protected_name = "_" + name
+
+    def __get__(self, instance: callable, owner: IntegerRange) -> None:
+        getattr(instance, self._protected_name)
+
+    def __set__(self, instance: callable, value: int) -> None:
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError
```
ArsenPidhoretskyi, line 19: Provide some description for this

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #52, opened 2023-01-21
'Solution'

StinGER-ua:


app/main.py
```diff
@@ -2,24 +2,62 @@
...
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: object, objtype: object) -> int:
+        return getattr(obj, self.protected_name)
+
+    def __set__(self, obj: object, value: int) -> None:
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError
```
Vasyl-Smutok, line 17: what's about other way?

HINT:
use `setattr`

Vasyl-Smutok (changes requested):

app/main.py
StinGER-ua: Need to add 
setattr(obj, self.protected_name, value) 
?

olena-hul (approved):

olena-hul:
Good job!

PR #57, opened 2023-01-29
Solution

ilushasm:


app/main.py
```diff
@@ -1,25 +1,70 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        check = self.limitation_class(
+            age=visitor.age,
+            weight=visitor.weight,
+            height=visitor.height
+        )
+        return all([check.age, check.weight, check.height])
```
Vasyl-Smutok, line 70: use `try: .... except`  here

```suggestion
        try:
             self.limitation_class(
                 visitor.age,
                 visitor.height,
                 visitor.weight
             )
         except (ValueError, TypeError):
             return False
         return True
```

Vasyl-Smutok (changes requested):

app/main.py
ilushasm: Okay, will do. However, it was not mentioned in ReedMe that it has to be handled with `try: ... except:`
What are the benefits of using `try: ... except:` here in comparison with original solution?

katevasilenko (approved):
Looks good to me

PR #69, opened 2023-02-07
Solution

Andrii-Kozlovskyi:


app/main.py
```diff
@@ -2,24 +2,66 @@
...
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: object) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
```
katevasilenko, line 15: You also can check here if value is integer and raise TypeError

katevasilenko (approved):
Looks good to me. It would be great if add additional checks

PR #72, opened 2023-02-10
'Solution'

natalie-goriela:


app/main.py
```diff
@@ -1,25 +1,119 @@
...
+    def set_age(self, value: int) -> None:
+        pass
+
+    @abstractmethod
+    def set_weight(self, value: int) -> None:
+        pass
+
+    @abstractmethod
+    def set_height(self, value: int) -> None:
+        pass
```
ArsenPidhoretskyi, line 63: No needed this

app/main.py
```diff
@@ -1,25 +1,119 @@
...
+    height = IntegerRange(80, 120)
+
+    def set_age(self, value: int) -> None:
+        self.age = value
+
+    def set_weight(self, value: int) -> None:
+        self.weight = value
+
+    def set_height(self, value: int) -> None:
+        self.height = value
```
ArsenPidhoretskyi, line 78: No needed

app/main.py
```diff
@@ -1,25 +1,119 @@
...
+    height = IntegerRange(120, 220)
+
+    def set_age(self, value: int) -> None:
+        self.age = value
+
+    def set_weight(self, value: int) -> None:
+        self.weight = value
+
+    def set_height(self, value: int) -> None:
+        self.height = value
```
ArsenPidhoretskyi, line 93: No needed

app/main.py
```diff
@@ -1,25 +1,119 @@
...
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator = None
```
ArsenPidhoretskyi, line 100: limitation_class is not instance of SlideLimitationValidator it is type[SlideLimitationValidator]
Also never None

app/main.py
```diff
@@ -1,25 +1,119 @@
...
+                other.age,
+                other.weight,
+                other.height
+            )
+        else:
+            self.limitation_class = AdultSlideLimitationValidator(
+                other.age,
+                other.weight,
+                other.height
+            )
```
ArsenPidhoretskyi, line 117: So bad practice to check by name. 
Example: I want to name my slide 'AdultSlide' but with Validator for Children

app/main.py
```diff
@@ -1,25 +1,119 @@
...
+    ) -> None:
+        return getattr(instance, self.protected_name)
+
+    def __set__(
+            self,
+            instance: SlideLimitationValidator,
+            value: int
+    ) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            return setattr(instance, self.protected_name, value)
```
ArsenPidhoretskyi, line 30: Otherwise raise ValueError

ArsenPidhoretskyi (changes requested):

app/main.py
natalie-goriela: Why do we need to inherit from ABC then?

app/main.py
natalie-goriela: Added ValueError, although tests crush in such case

app/main.py
```diff
@@ -1,25 +1,103 @@
...
+        # else:
+        #     raise ValueError(
+        #         f"{self.public_name} should be >= "
+        #         f"{self.min_amount} and <= {self.max_amount}")
```
olena-hul, line 35: Raise ValueError but without using variables in f-strings. Then tests should go ok

app/main.py
```diff
@@ -1,25 +1,103 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, other: Visitor) -> bool:
+        if self.limitation_class == ChildrenSlideLimitationValidator:
+            self.limitation_class = ChildrenSlideLimitationValidator(
+                other.age,
+                other.weight,
+                other.height
+            )
```
olena-hul, line 95: just self.limitation_class(other.age, other.weight, other.height)

app/main.py
```diff
@@ -1,25 +1,103 @@
...
+        else:
+            self.limitation_class = AdultSlideLimitationValidator(
+                other.age,
+                other.weight,
+                other.height
+            )
+
+        return True if len(self.limitation_class.__dict__) == 3 else False
```
olena-hul, line 103: Should return True if exception was not risen, False - otherwise

olena-hul (changes requested):

app/main.py
```diff
@@ -1,25 +1,92 @@
...
+        return getattr(instance, self.protected_name)
+
+    def __set__(
+            self,
+            instance: SlideLimitationValidator,
+            value: int
+    ) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            return setattr(instance, self.protected_name, value)
+        else:
```
ArsenPidhoretskyi, line 32: else is Redundant here

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #77, opened 2023-02-11
Solution

SafonovVladimir:
version 1

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,24 +2,74 @@ from abc import ABC
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int = None, max_amount: int = None) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: str, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: str, value: int) -> None:
+        return getattr(obj, self.protected_name)
+
+    def __set__(self, obj: str, value: int) -> (None, bool):
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(obj, self.protected_name, value)
+        else:
+            return False
 
 
 class Visitor:
-   pass
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,25 +1,65 @@
...
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: (ChildrenSlideLimitationValidator,
```
AndereLion, line 52: U can use щту parent SlideLimitationValidator class instead of  two child.

AndereLion (changes requested):

app/main.py
SafonovVladimir: Done

AndereLion (approved):
Super!

bihunva (approved):
Well done!

app/main.py
```diff
@@ -1,25 +1,64 @@
...
+
+    def __set_name__(self, owner: SlideLimitationValidator, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: SlideLimitationValidator, value: int) -> None:
+        return getattr(obj, self.protected_name)
+
+    def __set__(self, obj: str, value: int) -> None:
+        if value not in range(self.min_amount, self.max_amount + 1):
+            raise ValueError
```
Nattalli, line 18: it is better to specify the error with an additional message

app/main.py
```diff
@@ -1,25 +1,64 @@
...
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        age = visitor.age
+        height = visitor.height
+        weight = visitor.weight
+        try:
+            self.limitation_class(age, height, weight)
+            return True
+        except ValueError:
```
Nattalli, line 63: this error as well

Nattalli (approved):
Missclick, good job!

PR #79, opened 2023-02-12
Solution

bihunva:


NEkropo1 (changes requested):
image
I guess it shouldn't work that way, mb you need a little deeper slide inside hood, and refactor code according to readme )

app/main.py
```diff
@@ -63,6 +72,6 @@ def can_access(self, visitor: Visitor) -> bool:
                 height=visitor.height,
                 age=visitor.age
             )
-        except ValueError:
+        except (ValueError, TypeError):
```
NEkropo1, line 75: Guess python will handle rising TypeError, so you can leave only ValueError )

NEkropo1 (changes requested):

NEkropo1 (approved):
GJ!

app/main.py
```diff
@@ -1,25 +1,77 @@
...
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(
+                weight=visitor.weight,
+                height=visitor.height,
+                age=visitor.age
+            )
+        except ValueError:
+            return False
+        return True
```
Nattalli, line 77: you can `return True` in the `try` block

Nattalli (changes requested):

Nattalli (approved):
Good job!

PR #84, opened 2023-02-14
solution

amber-marichi:


kuznetsb (approved):
Good job!

oksana-feshchenko (approved):

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+    def __get__(
+        self,
+        instance: object,
+        owner: type[object]
+    ) -> int:
+        return getattr(instance, self.private_name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
+        assert self.min_amount <= value <= self.max_amount
+        setattr(instance, self.private_name, value)
```
Serhii-Leonenko, line 27: I think it's better to raise here `ValueError`. Also you can add validation for type of the value and raise TypeError

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.max_amount = max_amount
+        self.min_amount = min_amount
+
+    def __set_name__(
+        self,
+        owner: type[object],
+        name: str
+    ) -> None:
+        self.public_name = name
```
Serhii-Leonenko, line 15: As for me, you don't need a public_name attribute here, so you can remove it but it isn't obligatory.

Serhii-Leonenko (changes requested):

wQuelS (approved):
Nice

PR #85, opened 2023-02-15
Solution

kuznetsb:


Maytreyaya (approved):

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        if isinstance(self.limitation_class, AdultSlideLimitationValidator):
+            self.limitation_class = AdultSlideLimitationValidator
+        elif isinstance(self.limitation_class,
+                        ChildrenSlideLimitationValidator):
+            self.limitation_class = ChildrenSlideLimitationValidator
```
YuriiKindrat, line 63: why do u need this code? limitation_class is allready passed in this attribute

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: int) -> None:
+        if value not in range(self.min_amount, self.max_amount + 1):
+            return
```
YuriiKindrat, line 18: u can raise error here, so u can catch in in can access method

YuriiKindrat (changes requested):

app/main.py
kuznetsb: We receive class instance to the init of Slide. When I initialize self.limitation_class with parameters I like initialize new instance from instance? or how to understand it correctly. Also when I write self.limitation_class(visitor.age, visitor.weight, visitor.height) Pycharm underline me that it's not correct

wQuelS (approved):
Good

PR #89, opened 2023-02-15
Solution

oksana-feshchenko:


ZabFTFT (approved):
Nothing to add! Well done!

app/main.py
```diff
@@ -2,24 +2,63 @@
...
+        self.limitation_class = limitation_class
+
+    def can_access(self, obj: Visitor) -> bool:
+        try:
+            self.limitation_class(
+                age=obj.age, height=obj.height, weight=obj.weight
+            )
+        except ValueError:
+            return False
+        else:
```
YuriiKindrat, line 63: no need to use else here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -2,24 +2,62 @@
...
+    def __set_name__(self, owner: type[object], name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type[object]) -> int:
+        value = getattr(instance, self.protected_name)
+        return value
+
+    def __set__(self, instance: object, value: int) -> ValueError | None:
+        if value < self.min_amount or value > self.max_amount:
+            raise ValueError
```
YuriiKindrat, line 18: it is better to add some message to exception it is a bad practice to use exceptions without message

YuriiKindrat (approved):

PR #94, opened 2023-02-16
solution

MaxiTDI:


app/main.py
```diff
@@ -2,24 +2,82 @@
...
+    def is_valid(self, visitor: Visitor) -> bool:
+        value = [
+            self.age.min_amount <= visitor.age,
+            visitor.age <= self.age.max_amount,
+            self.weight.min_amount <= visitor.weight,
+            visitor.weight <= self.weight.max_amount,
+            self.height.min_amount <= visitor.height,
+            visitor.height <= self.height.max_amount
+        ]
+        return all(value)
```
ArsenPidhoretskyi, line 50: Redundant method

app/main.py
```diff
@@ -2,24 +2,82 @@
...
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+
+    def __init__(self) -> None:
+        super().__init__(
+            age=IntegerRange(4, 14),
+            weight=IntegerRange(20, 50),
+            height=IntegerRange(80, 120)
+        )
```
ArsenPidhoretskyi, line 60: Declare this as class attributes:

app/main.py
```diff
@@ -2,24 +2,82 @@
...
+            height=IntegerRange(120, 220)
+        )
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi, line 77: It accept type not instance

app/main.py
```diff
@@ -2,24 +2,82 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_validator = limitation_class()
```
ArsenPidhoretskyi, line 80: No need to create instance here

app/main.py
```diff
@@ -2,24 +2,82 @@
...
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_validator.is_valid(visitor)
```
ArsenPidhoretskyi, line 83: Use try-except statement
And remove is_valid method

app/main.py
ArsenPidhoretskyi: This check already present in IntegerRange

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @MaxiTDI not fixed)

ArsenPidhoretskyi: type[SlideLimitationValidator]

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #95, opened 2023-02-16
Bad solution

StreetNik:
I know the solution is bad, but I couldn't come up with a better solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,25 +1,76 @@
 from abc import ABC
+from typing import Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> str | int | bool:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> None:
+        setattr(instance, self.protected_name, value)
+        if not self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, False)
 
 
 class Visitor:
-   pass
+    age = IntegerRange(4, 60)
+    weight = IntegerRange(20, 120)
+    height = IntegerRange(80, 220)
+
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
+        self._name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,25 +1,73 @@
...
+        self.name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> str | int | bool:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> None:
+        setattr(instance, self.protected_name, value)
+        if not self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, False)
```
ArsenPidhoretskyi, line 20: raise Exception if not otherwise setattr
No need to always setattr

app/main.py
```diff
@@ -1,25 +1,73 @@
...
 
 
 class Visitor:
-   pass
+    age = IntegerRange(4, 60)
+    weight = IntegerRange(20, 120)
+    height = IntegerRange(80, 220)
```
ArsenPidhoretskyi, line 26: No need this inside Visitor

app/main.py
```diff
@@ -1,25 +1,73 @@
...
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    def check_access(self) -> bool:
+        return all([self.age, self.weight, self.height])
```
ArsenPidhoretskyi, line 42: No need this

app/main.py
```diff
@@ -1,25 +1,73 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi, line 61: It accept type, not instance
type[SlideLimitationValidator]

app/main.py
```diff
@@ -1,25 +1,73 @@
...
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return (
+            self.limitation_class(
+                visitor.age,
+                visitor.weight,
+                visitor.height
+            ).check_access()
+        )
```
ArsenPidhoretskyi, line 73: Use try-except

ArsenPidhoretskyi (changes requested):

app/main.py
StreetNik: Thank you, it's much easier to solve it with exceptions. Just because exceptions are the next topic I didn't realize how to solve this task better than I did it before

ArsenPidhoretskyi (approved):
Well done!

PR #97, opened 2023-02-17
Solution

WalDragnir:


app/main.py
```diff
@@ -1,25 +1,71 @@
...
-   pass
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
     pass
```
ArsenPidhoretskyi, line 37: Add init here

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator()) -> None:
```
ArsenPidhoretskyi, line 55: It is type not instance. Also you create Validator here

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+        self.name = name
+        self.limitation_class = limitation_class()
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class.age = visitor.age
+            self.limitation_class.weight = visitor.weight
+            self.limitation_class.height = visitor.height
+        except TypeError as t:
```
ArsenPidhoretskyi, line 64: You can use both
except TypeError, ValueError:

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+            print(f"Error in validation: {t}")
+            return False
+        except ValueError as v:
+            print(f"Error in validation: {v}")
+            return False
+        else:
```
ArsenPidhoretskyi, line 70: else is redundant here

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+                 name: str,
+                 limitation_class: SlideLimitationValidator()) -> None:
+        self.name = name
+        self.limitation_class = limitation_class()
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class.age = visitor.age
+            self.limitation_class.weight = visitor.weight
+            self.limitation_class.height = visitor.height
```
ArsenPidhoretskyi, line 63: self.limitation_class it is type do:
self.limitation_class(and pass params here)

ArsenPidhoretskyi (changes requested):

app/main.py
WalDragnir: Serhii Derevianko say this is error in readme, with init didnt work

app/main.py
WalDragnir: image
Dont work

vasylhnatiuk (approved):
Good job!

PR #109, opened 2023-03-03
Solution

Rename925:


app/main.py
```diff
@@ -1,25 +1,61 @@
...
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: Callable) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, other: Visitor) -> bool:
+        self.name = other.name
```
ArsenPidhoretskyi, line 56: Why you need this?
Remove this code

app/main.py
```diff
@@ -1,25 +1,61 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: Callable) -> None:
```
ArsenPidhoretskyi, line 51: limitation_class: Callable - is wrong it accepts type of SlideLimitationValidator

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,25 +1,64 @@
...
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
vasylhnatiuk, line 54: would be better to use`type(SlideLimitationValidator)` instead of SlideLimitationValidator

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job!

PR #124, opened 2023-03-30
Solution

goldenuni:


app/main.py
```diff
@@ -1,25 +1,71 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self, age: int, height: int, weight: int) -> None:
+        super().__init__(age=age, height=height, weight=weight)
```
EdAlekseiev, line 45: For what do you override `__init__`? I think it is unnecessary, you can delete it.

app/main.py
```diff
@@ -1,25 +1,71 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, height: int, weight: int) -> None:
+        super().__init__(age=age, height=height, weight=weight)
```
EdAlekseiev, line 54: For what do you override `__init__`? I think it is unnecessary, you can delete it.

app/main.py
```diff
@@ -1,25 +1,71 @@
...
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: Callable) -> None:
```
EdAlekseiev, line 58: `limitation_class` is always the subclass of SlideLimitationValidator class. You must change typehint.

EdAlekseiev (changes requested):

app/main.py
goldenuni: image
what I have to do with this?

EdAlekseiev: You must use `type[SlideLimitationValidator]` instead of `SlideLimitationValidator` as type hint in this case.

app/main.py
ArsenPidhoretskyi: @goldenuni 
This is not fixed. It is redundant `__init__`. It will call it by default

ArsenPidhoretskyi: Check in other places

app/main.py
ArsenPidhoretskyi: It accepts type of SlideLimitationValidator
Please fix typing
@goldenuni

ArsenPidhoretskyi (changes requested):

app/main.py
goldenuni: thanks)

app/main.py
```diff
@@ -2,24 +2,65 @@
...
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: [SlideLimitationValidator]) -> None:
```
Bohdan-Salamakha: u don't need to place class into parentheses

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #134, opened 2023-04-10
Solution

bezhevets:


vsmutok (approved):
Good Job 👍

app/main.py
```diff
@@ -1,25 +1,66 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: type(SlideLimitationValidator)
```
vadim-ilchenko: Use `Type[SlideLimitationValidator]` from `typing` module instead of `type(SlideLimitationValidator)` for type hinting.

app/main.py
```diff
@@ -1,25 +1,66 @@
...
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: str, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: str) -> str:
```
vadim-ilchenko: The return type should be `int` instead of `str` for the `__get__` method.

vadim-ilchenko (commented):
.

PR #135, opened 2023-04-12
Code for "Aquapark"

Niki-Alex:


app/main.py
```diff
@@ -1,25 +1,64 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(120, 220)
+    height = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: Callable) -> None:
```
ArsenPidhoretskyi: limitation_class accept type of SlideLimitationValidator

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,24 +2,66 @@
...
+            self,
+            name: str,
+            limitation_class: type(SlideLimitationValidator)
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.height, visitor.weight)
```
vasylhnatiuk: You should pass the visitor's attributes to the limitation_class constructor, like 'self.limitation_class(visitor.age, visitor.weight, visitor.height)' instead of 'self.limitation_class(visitor.age, visitor.height, visitor.weight)'.

app/main.py
```diff
@@ -2,24 +2,66 @@
...
+    weight = IntegerRange(120, 220)
+    height = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: type(SlideLimitationValidator)
```
vasylhnatiuk: Replace 'type(SlideLimitationValidator)' with 'Type[SlideLimitationValidator]' and add 'from typing import Type' at the beginning of the file for proper typing.

vasylhnatiuk (changes requested):

app/main.py
Niki-Alex: Swapped it, now it gives an error

app/main.py
```diff
@@ -1,25 +1,72 @@
...
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: object) -> str:
```
vsmutok: The return type annotation for the __get__ method should be `int`, not `str`. Change the return type to `int`.

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #136, opened 2023-04-13
Solution

anastasia-martyniuk:


app/main.py
```diff
@@ -1,25 +1,91 @@
...
+    ) -> None:
+        super().__init__(age, weight, height)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: It accepts type not instance

ArsenPidhoretskyi (changes requested):

app/main.py
anastasia-martyniuk: image
like this?

app/main.py
```diff
@@ -1,25 +1,91 @@
...
+    ) -> None:
+        super().__init__(age, weight, height)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: type(SlideLimitationValidator)
```
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

ArsenPidhoretskyi (changes requested):
Small mistake

app/main.py
ArsenPidhoretskyi: Yeah but type[And here class]

app/main.py
anastasia-martyniuk: I am using python 3.9 and here I can`t use type[], only Type[], is it ok?

ArsenPidhoretskyi (approved):
Well done!

PR #157, opened 2023-05-08
Solution

destr34:


app/main.py
```diff
@@ -1,25 +1,85 @@
 from abc import ABC
 
 
+class SlideLimitationValidator(ABC):
+    def __init__(self,
+                 age: int,
+                 weight: (int, float),
```
olena-hul:
```suggestion
                 weight: int | float,
```
Or Union can be used

app/main.py
```diff
@@ -1,25 +1,85 @@
...
 class IntegerRange:
-    pass
+    def __init__(self,
+                 min_amount: int,
+                 max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
 
+    def __set_name__(self,
+                     owner: SlideLimitationValidator,
```
olena-hul: Wrong annotation for owner. Fix in all places

app/main.py
olena-hul: Fix in all places

olena-hul (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #162, opened 2023-05-12
'Solution'

PetroHavrysh:


app/main.py
```diff
@@ -2,24 +2,60 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
```
ArsenPidhoretskyi: provide an inner type for limitation_class: type

app/main.py
```diff
@@ -2,24 +2,60 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.height, visitor.weight)
+        except ValueError:
+            return False
+        else:
+            return True
```
```suggestion
        return True
```
else is not needed

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #169, opened 2023-05-23
Solution

OksanaAdamchuk:


app/main.py
```diff
@@ -1,25 +1,71 @@
...
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: Any, objtype: Any = None) -> int:
+        return getattr(obj, self.protected_name)
+
+    def __set__(self, obj: Any, value: int) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(obj, self.protected_name, value)
+        else:
```
ArsenPidhoretskyi, line 19: else: is redundant in this case

app/main.py
```diff
@@ -1,25 +1,71 @@
...
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: Any, objtype: Any = None) -> int:
```
ArsenPidhoretskyi: objtype: type 
Also it does not have default value

app/main.py
```diff
@@ -1,25 +1,71 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age=age, weight=weight, height=height)
```
ArsenPidhoretskyi: Redundant code. It works automatically
The same in AdultSlideLimitationValidator

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
 
 
 class Slide:
-    pass
+    def __init__(
+            self, name: str, limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: it accepts type, not instance

ArsenPidhoretskyi (changes requested):

app/main.py
OksanaAdamchuk: but if I delete else block or just "else", code doesn't work any more.

Grayder0152 (approved):

PR #170, opened 2023-05-24
Develop

IvanKorshunovE:


app/main.py
```diff
@@ -1,25 +1,84 @@
...
+
+    def __get__(self,
+                instance: object,
+                owner: Type[object]) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object,
+                value: Any) -> Any:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, value)
```
Grayder0152, line 29: better to raise exception if value is incorrect
after that you can catch this exception in `can_access`

app/main.py
```diff
@@ -1,25 +1,84 @@
 from abc import ABC
+from typing import Type, Any
```
Grayder0152: sort imports

Grayder0152 (changes requested):

app/main.py
IvanKorshunovE: Got it! Thanks!

app/main.py
IvanKorshunovE: Screenshot 2023-05-24 at 20 14 17
Thanks for the feedback. Do you mean that I should include a blank line between the imports, similar to what is shown in the screenshot?

IvanKorshunovE: I re-checked order of imports lesson and can't figure out what is wrong with these imports:
Screenshot 2023-05-24 at 20 20 52

IvanKorshunovE: Screenshot 2023-05-24 at 20 23 15

Do you mean to sort all of them in alphabetical order?

vsmutok (approved):
Great 👍

PR #171, opened 2023-05-26
Aqua-park task complete

thorium-zero:
Yeah, I know about me using  that `Exception` class is kinda weird and stuff, but there are no instructions on how to handle exceptions for this project, so... :)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,25 +1,74 @@
 from abc import ABC
+from typing import Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.public_name = name
+        self.private_name = "_" + name
+
+    def __get__(self, obj: Any, objtype: Any = None) -> int:
+        return getattr(obj, self.private_name)
+
+    def __set__(self, obj: Any, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError("Value must be of type int")
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError("Invalid value")
+        setattr(obj, self.private_name, value)
 
 
 class Visitor:
-   pass
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,25 +1,74 @@
...
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__()
```
Oleg-s-s: Why you need this?

app/main.py
```diff
@@ -1,25 +1,74 @@
...
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator) -> None:
+        self.name: str = name
+        self.limitation_class: SlideLimitationValidator = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        result = True
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except Exception:
```
Oleg-s-s: I saw your comment but still using `Exception` directly can lead to missing some important ones. I think `ValueError` here will be enought

Oleg-s-s (changes requested):

app/main.py
thorium-zero: Yeah, I know, will fix in the next commit

app/main.py
```diff
@@ -1,25 +1,73 @@
...
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    def is_valid(self) -> bool:
+        pass
```
YuriiKindrat: why do u need this ?)

app/main.py
```diff
@@ -1,25 +1,73 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
YuriiKindrat: no need to rewrite init methods without changes

YuriiKindrat (changes requested):

app/main.py
thorium-zero: I don't, I though I will, though. And then I forgot to remove it

YuriiKindrat (approved):

TPolina (approved):

PR #173, opened 2023-05-28
Solution

yuro740:


app/main.py
```diff
@@ -1,25 +1,78 @@
...
+        self.max_amount = max_amount
+
+    def validate(self, value: int | float) -> None:
+        if not isinstance(value, int | float):
+            raise TypeError("Quantity should be int or float.")
+
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError(
+                f"Quantity should not be less than {self.min_amount}"
+                f"and greater than {self.max_amount}.")
```
vsmutok: I don't think you should create 2 classes, you can write everything in IntegerRange

vsmutok (changes requested):

app/main.py
yuro740: Why I must create all in one class? in prevoius task i also created 2 classes Validator and Number. But this is posible to create other validete for namber and OneOf. I think if we used descriptors in previus task and in this task. I can use it.
I wrote onetime validator and use it in other task?  just add float check.

yuro740: > I don't think you should create 2 classes, you can write everything in IntegerRange

Why I must create all in one class? in prevoius task i also created 2 classes Validator and Number. But this is posible to create other validate for Number and OneOf. I think if we used descriptors in previous task and in this task. I can use it.
I wrote onetime validator and use it in other task? just add "float" check.

ArsenPidhoretskyi (approved):
Well done!

Abnormaltype (approved):

PR #183, opened 2023-06-06
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1,25 +1,109 @@
...
+    ) -> bool:
+        return getattr(instance, self.protected_name)
+
+    def __set__(
+            self,
+            instance: Union[
+                "ChildrenSlideLimitationValidator",
+                "AdultSlideLimitationValidator"
+            ],
+            value: int
```
YuriiKindrat: could u use parent class for type annotation ?

app/main.py
```diff
@@ -1,25 +1,109 @@
...
+    ) -> None:
+        if not (self.min_amount <= value <= self.max_amount):
+            value = False
+        else:
+            value = True
+
+        setattr(instance, self.protected_name, value)
```
YuriiKindrat: why do u need to rewrite a value ?)
u can just raise an exception when condition didn't pass

app/main.py
```diff
@@ -1,25 +1,109 @@
...
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: Union[
+                "ChildrenSlideLimitationValidator",
+                "AdultSlideLimitationValidator"
+            ]
```
YuriiKindrat: same is here

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done!

PR #184, opened 2023-06-06
'Solution'

ant-komarov:


app/main.py
```diff
@@ -1,25 +1,75 @@
...
+
+    def __get__(self, instance: Any, owner: Any) -> None:
+        value = getattr(instance, self.protected_name)
+        return value
+
+    def __set__(self, instance: Any, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError
+        if self.max_amount < value or value < self.min_amount:
+            raise ValueError
```
YuriiKindrat: it is better to add messages to exceptions when u are raising them

YuriiKindrat (changes requested):

YuriiKindrat (approved):

app/main.py
```diff
@@ -1,25 +1,76 @@
...
 class IntegerRange:
-    pass
+    def __init__(self,
+                 min_amount: int,
+                 max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.public_nane = name
```
Oleg-s-s: public_nane ))

PR #190, opened 2023-06-07
My solution

kovaliskoveronika:


app/main.py
```diff
@@ -2,24 +2,65 @@
...
+
+    def can_access(self, visitor: Visitor) -> bool:
+        limitation = self.limitation_class(visitor.age,
+                                           visitor.weight,
+                                           visitor.height)
+        if getattr(limitation, "age") \
+                and getattr(limitation, "weight") \
+                and getattr(limitation, "height"):
+            return True
+        return False
```
Oleg-s-s: You can use previously created validator here

app/main.py
```diff
@@ -2,24 +2,65 @@
...
+    def __get__(self, instance: any, owner: any) -> any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: any, value: any) -> any:
+        if value < self.min_amount or value > self.max_amount:
+            setattr(instance, self.protected_name, None)
+            return
+        if type(value) != int:
+            setattr(instance, self.protected_name, None)
+            return
```
Oleg-s-s: No need to set `None` you can simply raise error on this step

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -2,24 +2,63 @@
...
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age,
+                                  visitor.weight,
+                                  visitor.height)
+        except ValueError:
+            return False
+        else:
+            return True
```
Oleg-s-s: No need to use else block

Oleg-s-s (changes requested):

Oleg-s-s (approved):
good job) :+1:

PR #195, opened 2023-06-10
Solution

YKonoplyov:


app/main.py
```diff
@@ -1,25 +1,72 @@
...
+    def __set_name__(self, owner: Type, name: str) -> None:
+        self.name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: "Visitor", owner: Type) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: "Visitor", value: int) -> None:
+        if not (self.min_amount <= value <= self.max_amount):
+            raise VisitorChararacteristicError
```
YuriiKindrat, line 22: it is better to write some message with raising error

app/main.py
```diff
@@ -1,25 +1,72 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
YuriiKindrat, line 47: why do u need this here?

app/main.py
```diff
@@ -1,25 +1,72 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
YuriiKindrat, line 56: same

app/main.py
```diff
@@ -1,25 +1,72 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except VisitorChararacteristicError:
+            print("You don`t meet the conditions")
+            return False
+        return True
```
YuriiKindrat, line 72: could u move this return inside try-except block ?

YuriiKindrat (changes requested):

PR #205, opened 2023-06-19
Solution

sulym:


app/main.py
```diff
@@ -1,25 +1,76 @@
...
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: Type
```
YukihiroSM: Add more descriptive type definition. `Type` is too wide.

app/main.py
```diff
@@ -1,25 +1,76 @@
...
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: object) -> int:
```
YukihiroSM: change typehints. Object is too wide definition

YukihiroSM: Fix it everywhere

YukihiroSM (changes requested):

YukihiroSM (approved):
Adding Type[...] is not required. For example owner: "Visitor" should be enough

PR #206, opened 2023-06-21
Solution

jinjo0222988:


app/main.py
```diff
@@ -2,24 +2,75 @@
...
-    pass
+
+    def __set_name__(self, owner: any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: any, owner: any) -> int:
```
ArsenPidhoretskyi: instance -> object
owner -> type
value -> Any(capitalized!!!!)

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+        self.weight = weight
+        self.height = height
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(age=IntegerRange(4, 14),
+                         weight=IntegerRange(20, 50),
+                         height=IntegerRange(80, 120))
```
```suggestion
        age = IntegerRange(4, 14),
        weight = IntegerRange(20, 50)
        height = IntegerRange(80, 120)
```
The same fix in AdultSlideLimitationValidator

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+                         height=IntegerRange(120, 220))
 
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator
+                 ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class()
```
ArsenPidhoretskyi: Why you create it? Just same class

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+    def __init__(self) -> None:
+        super().__init__(age=IntegerRange(14, 60),
+                         weight=IntegerRange(50, 120),
+                         height=IntegerRange(120, 220))
 
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: It accepts type not instance

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+
+    def __height_access(self, visitor: Visitor) -> bool:
+        return self.limitation_class.height.min_amount \
+            <= visitor.height <= \
+            self.limitation_class.height.max_amount
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.__age_access(visitor) \
+            and self.__weight_access(visitor) \
+            and self.__height_access(visitor)
```
ArsenPidhoretskyi: In this case you can create limitation_class:
self.limitation_class(pass visitor params here)

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+
+    def __weight_access(self, visitor: Visitor) -> bool:
+        return self.limitation_class.weight.min_amount\
+            <= visitor.weight <= \
+            self.limitation_class.weight.max_amount
+
+    def __height_access(self, visitor: Visitor) -> bool:
+        return self.limitation_class.height.min_amount \
+            <= visitor.height <= \
+            self.limitation_class.height.max_amount
```
ArsenPidhoretskyi: Strange and duplicate code

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: any, owner: any) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: any, value: any) -> None:
+        setattr(instance, self.protected_name, value)
```
ArsenPidhoretskyi, line 21: You need validate here

ArsenPidhoretskyi (changes requested):

app/main.py
jinjo0222988: Hello ! thanks for review ! But i dont quite understand, hot to replace "any" with "Any"  ? 
When i do it, i have error:  

 def __set__(self, instance: Any, value: any) -> None:
NameError: name 'Any' is not defined


image

app/main.py
jinjo0222988: I don't understand, do I need to remove the constructor here or not?

app/main.py
ArsenPidhoretskyi: any - is function
Any - is typing
for usage Any:
from typing import Any

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
ArsenPidhoretskyi: Yeah, you do not need constructor just write 
age = IntegerRange(4, 14)
 and others

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,25 +1,66 @@
...
+        self.protected_name = "_" + name
+
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: object, owner: type) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> None:
```
ArsenPidhoretskyi: instance - object

app/main.py
```diff
@@ -1,25 +1,66 @@
...
+        if not (self.min_amount <= value <= self.max_amount):
+            setattr(instance, self.protected_name, False)
+            return
```
ArsenPidhoretskyi: raise ValueError(some description here)

app/main.py
```diff
@@ -1,25 +1,66 @@
...
+        setattr(instance, self.protected_name, True)
```
```suggestion
        setattr(instance, self.protected_name, value)
```

app/main.py
```diff
@@ -1,25 +1,66 @@
...
+                 ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        answer = self.limitation_class(visitor.age,
+                                       visitor.weight,
+                                       visitor.height)
+
+        return answer.age and answer.weight and answer.height
```
```suggestion
        try:
            self.limitation_class(visitor.age, visitor.weight,visitor.height)
            return True
        except ValueErorr:
             return False
```

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @jinjo0222988 not fixed

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: object, owner: type) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> None:
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError()
```
ArsenPidhoretskyi: Provide some description inside ()

app/main.py
```diff
@@ -1,25 +1,65 @@
...
-    pass
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: type
```
ArsenPidhoretskyi: Provide inner type:
type[SlideLimitationValidator]

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #210, opened 2023-06-23
solution

uzlss:


app/main.py
```diff
@@ -1,25 +1,74 @@
...
 
 
 class IntegerRange:
-    pass
+
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Callable, name: str) -> None:
```
ArsenPidhoretskyi: owner is type
instance is object

app/main.py
```diff
@@ -1,25 +1,74 @@
...
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Callable, owner: Callable) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Callable, value: Any) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_name, value)
```
ArsenPidhoretskyi: otherwise you need to raise an Exception

app/main.py
```diff
@@ -1,25 +1,74 @@
...
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: it accepts type of SlideLimitationValidator, not instance

app/main.py
```diff
@@ -1,25 +1,74 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, instance: Visitor) -> bool:
+        return len(
+            self.limitation_class(
+                instance.age,
+                instance.weight,
+                instance.height
+            ).__dict__) == 3
```
ArsenPidhoretskyi: use try-except after implement raise exception inside `__set__`

ArsenPidhoretskyi (changes requested):

app/main.py
uzlss: flake8 says that I should not
I would rather use "finally" cuz there will be "return" anyway

uzlss: no, it does not work :(

Grayder0152 (approved):

PR #213, opened 2023-06-26
Task solution.

laskarj:


app/main.py
```diff
@@ -1,25 +1,106 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: "Visitor") -> bool:
+        return self.limitation_class(
+            age=visitor.age,
+            weight=visitor.weight,
+            height=visitor.height
+        ).get_access()
```
Grayder0152: What if `__set__` raise `TypeError` ?

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,25 +1,109 @@
...
 class IntegerRange:
-    pass
+
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+        self.status_access = None
+
+    def __set_name__(self,
+                     owner: "SlideLimitationValidator",
```
```suggestion
                     owner: Type[SlideLimitationValidator],
```

vasylhnatiuk: change here and in all of code

vasylhnatiuk (changes requested):

app/main.py
laskarj: owner: Type[SlideLimitationValidator],
E   NameError: name 'SlideLimitationValidator' is not defined

laskarj: If I arrange the classes by spelling, it also does not work because the class is not created for typing, and if I rearrange it, it comes out in other classes.
I can write "str" instead of a string constant.

app/main.py
```diff
@@ -1,25 +1,110 @@
...
+            height: int
+    ) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+
+class SlideLimitationValidator(ABC):
+    @abstractmethod
```
ArsenPidhoretskyi: Not true remove this

app/main.py
```diff
@@ -1,25 +1,110 @@
...
+    def __get__(self,
+                instance: Type[SlideLimitationValidator],
+                owner: Type[SlideLimitationValidator]
+                ) -> bool:
+        return getattr(instance, self.status_access)
 
-class SlideLimitationValidator(ABC):
-    pass
+    def __set__(self,
+                instance: Type[SlideLimitationValidator],
```
ArsenPidhoretskyi: instance is not type
Fix in all places

app/main.py
```diff
@@ -1,25 +1,110 @@
...
+    weight = IntegerRange(20, 50)
+    height = IntegerRange(80, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    def get_access(self) -> bool:
+        return all([self.age, self.weight, self.height])
```
ArsenPidhoretskyi: Remove this code it is redundant

ArsenPidhoretskyi: The same in AdultSlideLimitationValidator

app/main.py
```diff
@@ -1,25 +1,110 @@
...
-    pass
+    def __set__(self,
+                instance: Type[SlideLimitationValidator],
+                value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError(
+                f"{self.status_access} should be integer,"
+                f" not a {type(value)}!"
+            )
+        setattr(
```
ArsenPidhoretskyi: validate that value in range from min_amount and max_amount, if no raise exception

app/main.py
```diff
@@ -1,25 +1,110 @@
...
+class SlideLimitationValidator(ABC):
+    @abstractmethod
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = weight
+
+    @abstractmethod
+    def get_access(self) -> bool:
+        pass
```
ArsenPidhoretskyi: This method is redundant

app/main.py
```diff
@@ -1,25 +1,110 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Type[Visitor]) -> bool:
+        return self.limitation_class(
+            age=visitor.age,
+            weight=visitor.weight,
+            height=visitor.height
+        ).get_access()
```
ArsenPidhoretskyi: Use try-except here

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #215, opened 2023-06-28
Finish task

U-Ra-All:


app/main.py
```diff
@@ -1,25 +1,110 @@
...
+            self,
+            min_amount: int,
+            max_amount: int,
+    ) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(
+            self,
+            owner: Any,
```
YukihiroSM: Owner shouldn't be Any, it's too wide. Add more descriptive typehint, please :)
Fix it everywhere.

app/main.py
```diff
@@ -1,25 +1,110 @@
...
+            name: str
+    ) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(
+            self,
+            obj: Any,
+            obj_type: Any = None
```
YukihiroSM: If the parameter can be None - use Optional. Also, Any is too wide. Please, take a look into previous comment.

YukihiroSM (changes requested):
Fix typehints. Anything else looks good to me.

app/main.py
U-Ra-All: Done

app/main.py
U-Ra-All: Done

Bohdan-Salamakha (approved):

PR #232, opened 2023-07-11
solution

vkleshko:


app/main.py
```diff
@@ -1,25 +1,67 @@
...
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: it accepts type of SlideLimitationValidator, not instance

ArsenPidhoretskyi (changes requested):

PR #257, opened 2023-08-15
Solution

OlegatorLE:


app/main.py
```diff
@@ -1,25 +1,66 @@
...
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+        self.public_name = name
+
+    def __get__(self, instance: Any, owner: Any) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: int) -> None:
+        if value < self.min_amount or value > self.max_amount:
```
Kroll410: it should be `and`

app/main.py
```diff
@@ -1,25 +1,66 @@
 from abc import ABC
+from typing import Any, Type
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
```
Kroll410, line 6: you can create method `in(...)` so you can validate, but it's up to you

Kroll410 (changes requested):

app/main.py
OlegatorLE: there cannot be "and" here! how can the value be less than the minimum and more than the maximum at the same time?

app/main.py
OlegatorLE: I didn't catch that point. What are you mean?

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __contains__(self, value: int) -> object:
+        return self.min_amount <= value <= self.max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+        self.public_name = name
```
Serhii-Leonenko, line 15: I think you don't need public_name attribute here)

Serhii-Leonenko (changes requested):

app/main.py
OlegatorLE: i cant agree with you. i use this name in 23 line.

Kroll410 (approved):

PR #269, opened 2023-08-17
Solution

denys-source:


app/main.py
```diff
@@ -1,25 +1,72 @@
 from abc import ABC
+from typing import Generic, Type, TypeVar
 
+Instance = TypeVar("Instance")
+Value = TypeVar("Value")
```
Serhii-Leonenko: For what purpose do you use it?) Is it necessary?)

app/main.py
```diff
@@ -1,25 +1,72 @@
...
-class IntegerRange:
-    pass
+
+class IntegerRange(Generic[Instance, Value]):
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount: int = min_amount
+        self.max_amount: int = max_amount
+
+    def __set_name__(self, _: Instance, name: str) -> None:
+        self.public_name: str = name
```
Serhii-Leonenko: where do you use the public_name attribute?

app/main.py
```diff
@@ -1,25 +1,72 @@
...
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+        self, name: str, limitation_class: Type[SlideLimitationValidator]
+    ) -> None:
+        self.name: str = name
```
Serhii-Leonenko: you already have have str annotation inside the parameters

Serhii-Leonenko (changes requested):

app/main.py
denys-source: `public_name` is used in the `ValueError` message

Kroll410 (approved):

PR #273, opened 2023-08-21
Solution

VasylTurok:


app/main.py
```diff
@@ -2,24 +2,96 @@
...
+        self.private_name = "_" + name
+        print(owner)
+
+    def __get__(self, instance: object, owner: object) -> float:
+        return getattr(instance, self.private_name)
+
+    def __set__(self, instance: object, value: float) -> float | None:
+        if self.min_amount <= value <= self.max_amount:
+            return setattr(instance, self.private_name, value)
+        print(f"Value must be"
```
pavlejviki: if the value is not in the allowed range better  raise ValueError with the appropriate message

app/main.py
```diff
@@ -2,24 +2,96 @@
...
 
 class Visitor:
-   pass
+
+    def __init__(
+            self,
+            name: str,
+            age: int | None,
+            weight: float | None,
+            height: float | None) -> None:
```
pavlejviki: None should not be acceptable for these parameters

app/main.py
```diff
@@ -2,24 +2,96 @@
...
 
 class SlideLimitationValidator(ABC):
-    pass
+
+    def __init__(
+            self,
+            name: str,
+            age: int | None,
+            weight: float | None,
+            height: float | None) -> None:
```
pavlejviki: None should not be acceptable for these parameters

app/main.py
```diff
@@ -2,24 +2,96 @@
...
+            self,
+            name: str,
+            limitation_class: AdultSlideLimitationValidator
+            | ChildrenSlideLimitationValidator
+    ) -> None:
+
+        self.limitation_class = limitation_class
+        self.name = name
+
+    def can_access(self, visitor: Visitor) -> bool:
```
pavlejviki, line 86: after following the above comments you can refactor this method taking into account that a ValueError can be raised. Consider using try/exept blocks

app/main.py
```diff
@@ -2,24 +2,96 @@
...
+    ) -> None:
+
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.private_name = "_" + name
+        print(owner)
+
+    def __get__(self, instance: object, owner: object) -> float:
```
pavlejviki: wrong type annotation for owner

app/main.py
```diff
@@ -2,24 +2,96 @@
...
 
 
 class Slide:
-    pass
+
+    def __init__(
+            self,
+            name: str,
+            limitation_class: AdultSlideLimitationValidator
+            | ChildrenSlideLimitationValidator
```
pavlejviki:
```suggestion
            limitation_class: SlideLimitationValidator
```

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,25 +1,92 @@
...
+        self.private_name = "_" + name
+
+    def __get__(self, instance: object, owner: Any) -> float:
+        return getattr(instance, self.private_name)
+
+    def __set__(self, instance: object, value: float) -> float | None:
+        if self.min_amount <= value <= self.max_amount:
+            return setattr(instance, self.private_name, value)
+        raise ValueError(f"Value must be"
+              f"{self.min_amount}...{self.max_amount}")
```
Dimosphen1: The lines of the f-string should be aligned with each other.

app/main.py
```diff
@@ -1,25 +1,92 @@
...
 
 class SlideLimitationValidator(ABC):
-    pass
+
+    def __init__(
+            self,
+            name: str,
+            age: int,
+            weight: float,
+            height: float) -> None:
```
Dimosphen1: Move `) -> None:` to a new line so that opening and closing parentheses will not have any arguments on their lines.

app/main.py
```diff
@@ -1,25 +1,92 @@
...
 
 class Visitor:
-   pass
+
+    def __init__(
+            self,
+            name: str,
+            age: int,
+            weight: float,
+            height: float) -> None:
```
Dimosphen1: Move `) -> None:` to a new line so that opening and closing parentheses will not have any arguments on their lines.

app/main.py
```diff
@@ -1,25 +1,92 @@
...
+        self.limitation_class = limitation_class
+        self.name = name
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            instance = self.limitation_class(visitor.name,
+                                             visitor.age,
+                                             visitor.weight,
+                                             visitor.height)
+            return True if instance else False
```
Dimosphen1: There is no need to obtain an `instance` here and add an if condition in the return statement, you could just call `self.limitation_class` with parameters and return True in case `ValueError` is not raised.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,25 +1,94 @@
...
+            name: str,
+            limitation_class: SlideLimitationValidator
+    ) -> None:
+
+        self.limitation_class = limitation_class
+        self.name = name
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.name,
```
Polyakiv-Andrey: After opening the parentheses, write all the arguments on a new line, including the first one, the closing bracket is also written on a new line

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #276, opened 2023-08-23
solution

EvgenMilokumov:


app/main.py
```diff
@@ -1,25 +1,63 @@
...
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: SlideLimitationValidator, owner: Any) -> None:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: Any) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_name, True)
+        else:
+            setattr(instance, self.protected_name, False)
```
ArsenPidhoretskyi: If value not in range you can raise an Exception

app/main.py
```diff
@@ -1,25 +1,63 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: Any) -> None:
```
ArsenPidhoretskyi: limitation_class is type

app/main.py
```diff
@@ -1,25 +1,63 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, instance: Visitor) -> bool:
+        limits = self.limitation_class(
+            instance.age, instance.weight, instance.height
+        )
+        if limits.age and limits.weight and limits.height:
+            return True
+        return False
```
ArsenPidhoretskyi: Not true use try-except after implement raise Exception

ArsenPidhoretskyi (changes requested):

app/main.py
EvgenMilokumov: done

app/main.py
EvgenMilokumov: done

app/main.py
EvgenMilokumov: done

app/main.py
```diff
@@ -1,25 +1,67 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
```
ArsenPidhoretskyi: type of what? Provide inner type

ArsenPidhoretskyi (changes requested):

app/main.py
EvgenMilokumov: done

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+    def __set_name__(self, owner: Any, name: Any) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: SlideLimitationValidator, owner: Any) -> None:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: Any) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_name, True)
```
ArsenPidhoretskyi: set value, no need to set True

ArsenPidhoretskyi:
```suggestion
            setattr(instance, self.protected_name, value)
```

ArsenPidhoretskyi (changes requested):

app/main.py
EvgenMilokumov: clear, thanks

ArsenPidhoretskyi (approved):
Well done!

PR #285, opened 2023-09-12
Solution

Filllick:
Pycharm highlights limitation_class: SlideLimitationValidator  and warns that 'SlideLimitationValidator' object is not callable, but the code passes all tests

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,4 +1,5 @@
 from abc import ABC
+from typing import Type
 
 
 class IntegerRange:
@@ -13,8 +14,9 @@ class IntegerRange:
         return getattr(instance, self.protected_name)
 
     def __set__(self, instance: callable, value: int) -> None:
-        if value in range(self.min_amount, self.max_amount + 1):
-            setattr(instance, self.protected_name, value)
+        if value not in range(self.min_amount, self.max_amount + 1):
+            raise ValueError
+        setattr(instance, self.protected_name, value)
 
 
 class Visitor:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,24 +2,58 @@
...
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator
+                 ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, client: Visitor) -> bool:
+        return len(self.limitation_class(
```
vasylhnatiuk: why do you use dict?it is strange logic. it would be better to use try-except here

app/main.py
```diff
@@ -2,24 +2,58 @@
...
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator
```
```suggestion
                 limitation_class: Type[SlideLimitationValidator]
```

app/main.py
```diff
@@ -2,24 +2,58 @@
...
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: callable, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: callable, owner: callable) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: callable, value: int) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
```
vasylhnatiuk: and raise here ValueError `if value not in range`

vasylhnatiuk (changes requested):

Filllick:
Added raise ValueError to IntegerRange.__set__, added try-except to Slide.can_access method

app/main.py
```diff
@@ -1,25 +1,62 @@
...
+                 ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, client: Visitor) -> bool:
+        try:
+            self.limitation_class(client.age, client.height, client.weight)
+            return True
+        except ValueError:
+            return False
```
vasylhnatiuk, line 61:
```suggestion
            return False
        return True
```

app/main.py
```diff
@@ -1,25 +1,62 @@
...
+                 name: str,
+                 limitation_class: Type[SlideLimitationValidator]
+                 ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, client: Visitor) -> bool:
+        try:
+            self.limitation_class(client.age, client.height, client.weight)
+            return True
```
vasylhnatiuk: and delete from here

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #291, opened 2023-09-20
Solution

Lebwa1337:


app/main.py
```diff
@@ -1,25 +1,81 @@
...
-    pass
+    def __init__(self) -> None:
+        super().__init__((4, 14), (20, 50), (80, 120))
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return (
+            self.age_range[0] <= visitor.age <= self.age_range[1]
+            and self.weight_range[0] <= visitor.weight <= self.weight_range[1]
+            and self.height_range[0] <= visitor.height <= self.height_range[1]
+        )
```
Dimosphen1: You should not define the `__init__` or any other methods in child classes of the `SlideLimitationValidator` class. Use class attributes where you will assign instances of the `IntegerRange` class with corresponding values. You won't need `can_access` logic then because you will rely on the `__set__` method of the `IntegerRange` class

app/main.py
```diff
@@ -1,25 +1,81 @@
...
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Dimosphen1: Change your type annotation to indicate that you expect to obtain subclasses of the `SlideLimitationValidator` class for `limitation_class`, not `SlideLimitationValidator` instances.

app/main.py
```diff
@@ -1,25 +1,81 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class()
```
Dimosphen1: There is no need to initialize the `limitation_class` in the `__init__`, it should be done in the `can_access` method with further error processing.

Dimosphen1 (changes requested):

app/main.py
Lebwa1337: but, number 5 says: 
Slide class
Its __init__ method takes 2 arguments:

name - string value, slide's name
limitation_class - type of SlideLimitationValidator class which sets restrictions on the use of the slide

Lebwa1337: but :
Its __init__ method takes 2 arguments:

name - string value, slide's name
limitation_class - type of SlideLimitationValidator class which sets restrictions on the use of the slide

viktoria-rybenchuk (approved):
Good job!

PR #292, opened 2023-09-20
implemented py-aquapark task

HlovaOleh:


app/main.py
```diff
@@ -1,25 +1,71 @@
...
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> object:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: Any) -> None:
```
olena-hul: Annotations for instance and owner are wrong

olena-hul (changes requested):

app/main.py
HlovaOleh: I'm trying to understand my mistake in the type annotations for instance and owner, but in my opinion, the types are correct

YukihiroSM: Instance is not just object, it's a specific object of class IntegerRange.  `Type[IntegerRange]`,

Owner is an object of Validation class: `Type[SlideLimitationValidator]`

YukihiroSM (changes requested):

app/main.py
HlovaOleh: I can not add Type[IntegerRange] annotation in class IntegerRange

YukihiroSM: You can add this type annotation if you import this: `from __future__ import annotations` at the beginning of the file. 
image

YukihiroSM (changes requested):

app/main.py
HlovaOleh: Got it, sorry about that

Bohdan-Salamakha (approved):

PR #295, opened 2023-09-24
'Solution'

TROTPOL:


app/main.py
```diff
@@ -2,24 +2,80 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: callable, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: callable, owner: callable) -> int:
```
Dimosphen1, line 13: It would be better to use `Callable` from `typing` which is [not deprecated for type hinting](https://docs.python.org/3/library/typing.html#annotating-callable-objects) than using `callable` directly.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #296, opened 2023-09-28
Solution

MatsalakO:


app/main.py
```diff
@@ -1,25 +1,94 @@
...
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+        self,
+        name: str,
+        limitation_class: SlideLimitationValidator
```
Dimosphen1: You expect to obtain a sub class of the `SlideLimitationValidator` class, not an instance of this class. Change your annotation to encounter that.

app/main.py
```diff
@@ -1,25 +1,94 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(
+                visitor.age,
+                visitor.weight,
+                visitor.height)
```
Dimosphen1: Place a closing parenthesis `)` on a new line so that you would not interrupt the code structure and will match the opening parenthesis `(` which does not have any arguments.

app/main.py
```diff
@@ -1,25 +1,94 @@
...
+        self,
+        owner: SlideLimitationValidator,
+        name: str
+    ) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(
+        self,
+        instance: IntegerRange,
+        owner: SlideLimitationValidator
```
Dimosphen1: You should not bind the `IntegerRange` class and its methods to the `SlideLimitationValidator` class instance in the type annotation. Replace your type annotations for `instance` and `owner` with `object` across the methods of the `IntegerRange` class.

Dimosphen1 (changes requested):

app/main.py
MatsalakO: > You should not bind the `IntegerRange` class and its methods to the `SlideLimitationValidator` class instance in the type annotation. Replace your type annotations for `instance` and `owner` with `object` across the methods of the `IntegerRange` class.

Can you help me in this point? Because, I really don`t understand, what i need to do

viktoria-rybenchuk: Using object as the type annotation is a common and practical approach for descriptors to work effectively with a wide range of classes

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
Well done!

PR #301, opened 2023-10-10
I don't know what type to write owner, obj, objtype - errors flake8

OlenaLis123:
I don't know what type to write owner, obj, objtype - errors flake8

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,25 +1,72 @@
 from abc import ABC
+from typing import Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: Any, objtype: None = None) -> int:
+        value = getattr(obj, self.protected_name)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,24 +2,70 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner, name: str) -> None:
```
```suggestion
    def __set_name__(self, owner: Any, name: str) -> None:
```

app/main.py
```diff
@@ -2,24 +2,70 @@
...
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, obj, objtype=None) -> int:
```
```suggestion
    def __get__(self, obj: Any, objtype=None) -> int:
```

app/main.py
```diff
@@ -2,24 +2,70 @@
...
+        value = getattr(obj, self.protected_name)
+        print(f"Accessing {self.public_name} giving {value}")
+        return value
+
+    def __set__(self, obj, value: int) -> bool:
```
```suggestion
    def __set__(self, obj: Any, value: int) -> bool:
```

EdAlekseiev (changes requested):

app/main.py
OlenaLis123: Дякую !

EdAlekseiev (approved):
Good job

PR #303, opened 2023-10-11
Solution

D1st3f:


YuliiaSaliuta (approved):

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.public_name = name
+        self.private_name = "_" + name
+
+    def __get__(self, obj: Any, objtype: Any = None) -> int:
+        return getattr(obj, self.private_name)
+
+    def __set__(self, obj: Any, value: int) -> None:
```
LiudmylaKulinchenko: better to use default parameters for these methods

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+    def __init__(self, name: str,
+                 limitation_class: Type[SlideLimitationValidator]) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        result = True
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except Exception:
```
LiudmylaKulinchenko: It is a bad practice to catch Exception

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+        self.private_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> float:
+        return getattr(instance, self.private_name)
+
+    def __set__(self, instance: object, value: float) -> None:
+        if not isinstance(value, int):
+            raise TypeError("Value must be of type int")
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError("Invalid value")
```
LiudmylaKulinchenko: You should describe here why is it invalid. It will be really hard to fix an error with such message

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        result = True
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
+            result = False
+        return result
```
LiudmylaKulinchenko: You can put return inside try

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job!

PR #305, opened 2023-10-12
Solution

YuliiaSaliuta:


Naz-iv (approved):
well done

app/main.py
```diff
@@ -2,24 +2,69 @@
...
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+
+        validator = self.limitation_class(
+            visitor.age,
+            visitor.weight,
+            visitor.height
+        )
+        return all(field for field in validator.__dict__.values())
```
LiudmylaKulinchenko: `Create can_access method that takes instance of Visitor class and returns if the visitor can use the slide.` -- you should have this logic here, not in setter

app/main.py
```diff
@@ -2,24 +2,69 @@
...
+
+    def __get__(self, instance: object, owner: type) -> str:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_name, value)
+        else:
+            setattr(instance, self.protected_name, False)
```
LiudmylaKulinchenko: raise a ValueError if value is in wrong range

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,24 +2,67 @@
...
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> str:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError(f"Value {value} not in range")
```
LiudmylaKulinchenko: Better to specify what a range is here, because it is not really clear now

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,24 +2,67 @@
...
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> str:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError(f"Value {value} not in range for this slide")
```
LiudmylaKulinchenko, line 19: You can also specify min and max amount

LiudmylaKulinchenko (approved):
Great! 😋

PR #306, opened 2023-10-12
Solution

Naz-iv:


app/main.py
```diff
@@ -1,25 +1,73 @@
...
+        self.protected_name = "_" + value
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_name, value)
+        else:
+            setattr(instance, self.protected_name, False)
```
Lyutillis: Here you can raise Exception instead of assigning False to variable

app/main.py
```diff
@@ -1,25 +1,73 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> Visitor:
+        validator = self.limitation_class(
+            visitor.age,
+            visitor.weight,
+            visitor.height
+        )
+        return all(field for field in validator.__dict__.values())
```
Lyutillis: Here you could use try\except statement if your __set__ method raised an Exception instead of checking every field for their value

Lyutillis (approved):

app/main.py
Naz-iv: nice, will update

app/main.py
Naz-iv: will do, thanks

vshvanska (approved):

app/main.py
```diff
@@ -1,25 +1,73 @@
...
 
 
 class IntegerRange:
-    pass
+
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, value: str) -> None:
```
LiudmylaKulinchenko: Don't use any here. It won't work with any type

app/main.py
```diff
@@ -1,25 +1,73 @@
...
+        self.protected_name = "_" + value
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
```
LiudmylaKulinchenko: same about instance. Fix everywhere

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: updated, though I'm not sure what type I should use since descriptors should handle any type

app/main.py
```diff
@@ -2,24 +2,71 @@
...
+
+    def __set_name__(self, owner: type, value: str) -> None:
+        self.protected_name = "_" + value
+
+    def __get__(self, instance: type, owner: type) -> type:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: type, value: type) -> None:
+        if value not in range(self.min_amount, self.max_amount + 1):
+            raise ValueError
```
fsocie7y: it is better to add some message to exception

fsocie7y (changes requested):

app/main.py
```diff
@@ -2,24 +2,71 @@
...
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> Visitor:
+        try:
+            self.limitation_class(visitor.age,
+                                  visitor.weight,
+                                  visitor.height)
+        except ValueError:
+            return False
+        return True
```
LiudmylaKulinchenko, line 72: you can move this return inside `try`

LiudmylaKulinchenko (approved):
Well done!

PR #308, opened 2023-10-12
Solution

kravetb:


app/main.py
```diff
@@ -1,25 +1,86 @@
...
-    pass
+    def __init__(
+        self,
+        min_amount: int,
+        max_amount: int
+    ) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Type, name: str) -> None:
```
```suggestion
    def __set_name__(self, owner: Any, name: str) -> None:
```

app/main.py
```diff
@@ -1,25 +1,86 @@
...
+        self.protected_name = "_" + name
+
+    def __get__(
+        self,
+        instance: Optional[Type],
```
EdAlekseiev: Why is it `Optional`?

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
Still not fixed

app/main.py
kravetb: How type I need import for instance?

EdAlekseiev (approved):
Good job

PR #310, opened 2023-10-12
py-aquapark Solution

RVChornyy:


app/main.py
```diff
@@ -2,24 +2,69 @@
...
+        self.max_amount = max_amount
+
+    def __get__(self,
+                instance: object,
+                owner: type) -> int | float:
+        return getattr(instance, self.proteted_name)
+
+    def __set__(self,
+                instance: object,
+                value: int | float) -> None:
```
prochigor: excepted value only int

app/main.py
```diff
@@ -2,24 +2,69 @@
...
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self,
+                instance: object,
+                owner: type) -> int | float:
+        return getattr(instance, self.proteted_name)
+
+    def __set__(self,
+                instance: object,
```
prochigor: annotation object is too plural, 'instanse' is instance of the class Visitor

app/main.py
```diff
@@ -2,24 +2,69 @@
...
-    pass
+    def __init__(self,
+                 min_amount: int,
+                 max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self,
+                instance: object,
+                owner: type) -> int | float:
```
prochigor: you can do it in one line, there aren't many words

prochigor (changes requested):

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+        self.max_amount = max_amount
+
+    def __get__(self, instance: Visitor, owner: type) -> int:
+        return getattr(instance, self.proteted_name)
+
+    def __set__(self,
+                instance: Visitor,
+                value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError
```
LiudmylaKulinchenko: You should add a message when raising error

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age,
+                                  visitor.weight,
+                                  visitor.height)
+        except ValueError:
+            return False
+        return True
```
LiudmylaKulinchenko: you can place this return iside `try`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+        self.max_amount = max_amount
+
+    def __get__(self, instance: Visitor, owner: type) -> int:
+        return getattr(instance, self.proteted_name)
+
+    def __set__(self,
+                instance: Visitor,
+                value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError("Visitor can't use the slide")
```
LiudmylaKulinchenko: You should use more descriptive message here. It is an descriptor, so you want to add a message that will describe that ValueError was raised because of your condition "if not self.min_amount <= value <= self.max_amount"

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Stunning!

PR #312, opened 2023-10-13
"Aqua park" task completed.

AndriiHomeniuk1:


app/main.py
```diff
@@ -1,25 +1,73 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(
+            self,
+            instance: Visitor,
+            owner: Type[Visitor]
+    ) -> Union[str, int]:
```
```suggestion
    ) ->Optional[Union[str, int]]:
```

vasylhnatiuk (changes requested):

app/main.py
AndriiHomeniuk1: I understood that's much better.
Added return annotation with Optional.

vasylhnatiuk (approved):

PR #313, opened 2023-10-14
Solution

prochigor:


YuliiaSaliuta (approved):

app/main.py
```diff
@@ -1,25 +1,76 @@
...
+    def __set_name__(self, owner: any, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Visitor, owner: any) -> str:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Visitor, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError
```
LiudmylaKulinchenko: it is better to write some message with raising error

app/main.py
```diff
@@ -1,25 +1,76 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
+            return False
+        return True
```
LiudmylaKulinchenko: You can move this return inside try/except block

LiudmylaKulinchenko (changes requested):

app/main.py
prochigor: Ok, but I always handle this raise in the function can_access

app/main.py
prochigor: Done

app/main.py
LiudmylaKulinchenko: But it also should work by itself

LiudmylaKulinchenko (approved):
Well done! 🔥🔥

PR #319, opened 2023-10-22
Solution

DanSheremeta:
Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,25 +1,96 @@
 from abc import ABC
+from typing import Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: Any) -> Any:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> Any:
+        if not isinstance(value, int):
+            raise TypeError()
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError()
+        setattr(instance, self.protected_name, value)
 
 
 class Visitor:
-   pass
+    def __init__(
+            self,
+            name: str,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(
+            self,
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,25 +1,96 @@
...
+from typing import Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: Any) -> Any:
```
EdAlekseiev: `name` is always `str`. Return type is `None`

```suggestion
    def __set_name__(self, owner: Any, name: str) -> None:
```

app/main.py
```diff
@@ -1,25 +1,96 @@
...
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
```
EdAlekseiev: Invalid type hint for return type

app/main.py
```diff
@@ -1,25 +1,96 @@
...
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> Any:
```
EdAlekseiev: Invalid type hint for return type

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,25 +1,96 @@
...
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: Any) -> None:
```
YukihiroSM: Value can be int only, fix typehint

app/main.py
```diff
@@ -1,25 +1,96 @@
...
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(
+            self,
+            age: int,
+            height: int,
+            weight: int
+    ) -> None:
+        super().__init__(age, height, weight)
```
YukihiroSM: this init is not needed  here

app/main.py
```diff
@@ -1,25 +1,96 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(
+            self,
+            age: int,
+            height: int,
+            weight: int
+    ) -> None:
+        super().__init__(age, height, weight)
```
YukihiroSM: this init is not needed here

YukihiroSM (changes requested):

YukihiroSM (approved):

PR #325, opened 2023-10-24
Solution

olgierrd:


app/main.py
```diff
@@ -1,25 +1,78 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Dimosphen1: You should indicate in type annotations that you expect not a `SlideLimitationValidator` instance for `limitation_class`, but rather a subclass of the `SlideLimitationValidator` class — change type annotation to achieve that.

Dimosphen1 (changes requested):
One minor change was requested.

app/main.py
```diff
@@ -1,25 +1,79 @@
...
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class:
+            AdultSlideLimitationValidator or ChildrenSlideLimitationValidator
```
Dimosphen1: In this way, you indicate that you expect an instance of `AdultSlideLimitationValidator` or  `ChildrenSlideLimitationValidator` class. Besides that, the common syntax for indicating or is using `|`. To indicate a subclass of certain class, you could use the following syntax:
```suggestion
            limitation_class: type[SlideLimitationValidator]
```
Apply this change to your code.

Dimosphen1 (changes requested):

pavlejviki (approved):
GJ!

PR #332, opened 2023-10-31
'Slution'

zakotii:


app/main.py
```diff
@@ -1,25 +1,66 @@
...
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: int, owner: int) -> None:
+        return instance.__dict__[self.name]
+
+    def __set__(self, instance: int, value: int) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            instance.__dict__[self.name] = value
```
ArsenPidhoretskyi: use setattr

app/main.py
```diff
@@ -1,25 +1,66 @@
...
+    age = IntegerRange(0, 150)
+    weight = IntegerRange(0, 500)
+    height = IntegerRange(0, 300)
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    @abstractmethod
+    def can_access(self, visitor: Visitor) -> bool:
+        pass
```
ArsenPidhoretskyi: This method is redundant in all places

app/main.py
```diff
@@ -1,25 +1,66 @@
...
-    pass
+    def __init__(
+            self, name: str,
+            limitation_class: SlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_class().can_access(visitor)
```
```suggestion
        return self.limitation_class(and pass here all parameters)
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,24 +2,67 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: int, owner: int) -> None:
+        return instance.__dict__[self.name]
```
EdAlekseiev: Use `getattr` here, please

EdAlekseiev (changes requested):

app/main.py
ArsenPidhoretskyi: not fixed

app/main.py
```diff
@@ -2,24 +2,64 @@
...
 
 class SlideLimitationValidator(ABC):
-    pass
+    def can_access(self, visitor: Visitor) -> bool:
+        pass
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def can_access(self, visitor: Visitor) -> bool:
```
ArsenPidhoretskyi: This method redundant. Remove it from all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
What about other comments? They are not fixed

EdAlekseiev (approved):
Good job

PR #333, opened 2023-10-31
Solution

spryima:


app/main.py
```diff
@@ -1,25 +1,61 @@
...
 
 
 class Slide:
-    pass
+
+    def __init__(self, name: str, limitation_class: Type) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool | str:
```
```suggestion
    def can_access(self, visitor: Visitor) -> bool:
```

vasylhnatiuk: it returns True or False

app/main.py
```diff
@@ -1,25 +1,61 @@
...
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+
+    def __init__(self, name: str, limitation_class: Type) -> None:
```
```suggestion
    def __init__(self, name: str, limitation_class: SlideLimitationValidator) -> None:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
do it

spryima:
when I use this code 
   def __init__(self, name: str, limitation_class: SlideLimitationValidator) -> None:
  i get 
  SlideLimitationValidator' object is not callable

  this one works without such problem:
  limitation_class: Type[SlideLimitationValidator]) -> None:

app/main.py
```diff
@@ -1,25 +1,63 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
+            return False
+        else:
+            return True
```
```suggestion
        return True
```

ArsenPidhoretskyi: else: is redundant in this case

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #335, opened 2023-11-02
Solution

BirovShandor:


app/main.py
```diff
@@ -2,24 +2,76 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator) -> None:
```
ArsenPidhoretskyi: it accepts type of,  not instance.
Use type[SlideLimitationValidator]

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
.

YukihiroSM (approved):
Nicely done! Thanks for fixing this!

PR #337, opened 2023-11-02
Solution

CrazyHilly:


app/main.py
```diff
@@ -1,25 +1,84 @@
...
+            min_amount: int,
+            max_amount: int
+    ) -> None:
+
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: SlideLimitationValidator, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
```
EdAlekseiev: Use f-string here

app/main.py
```diff
@@ -1,25 +1,84 @@
...
+
+    def __get__(
+            self,
+            instance: SlideLimitationValidator,
+            owner: SlideLimitationValidator
+    ) -> str:
```
EdAlekseiev: `str` is incorrect type hint for return type

app/main.py
```diff
@@ -1,25 +1,84 @@
...
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
+    def validate(self) -> bool:
+        pass
```

app/main.py
```diff
@@ -1,25 +1,84 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def validate(self) -> bool:
+        return True if self.age and self.height and self.weight else False
```

app/main.py
```diff
@@ -1,25 +1,84 @@
...
+    ) -> None:
+
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        age = visitor.age
+        weight = visitor.weight
+        height = visitor.height
+        return self.limitation_class(age, weight, height).validate()
```
```suggestion
        return self.limitation_class(age, weight, height)
```

EdAlekseiev (changes requested):

app/main.py
CrazyHilly: I used this method because I set attribute value to 0 in IntegerRange class if value was out of range, so I needed to catch it later. I couldn't think of any other way. 

I could have used: 
return (self.limitation_class.age == visitor.age and
            self.limitation_class.height == visitor.height and
            self.limitation_class.weight == visitor.weight)
in can_access method directly but it doesn't look nice.

I now replaced it with exception handling, although it's from the following topics as I don't have any other ideas. If there is another way, please let me know.

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
EdAlekseiev: Type hint for `limitation_class`  is incorrect. You should use `Type[SlideLimitationValidator]`, because limitation_class is not of instance `SlideLimitationValidator `

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,25 +1,79 @@
...
+
+    def can_access(self, visitor: Visitor) -> bool:
+        age = visitor.age
+        weight = visitor.weight
+        height = visitor.height
+        try:
+            self.limitation_class(age, weight, height)
+        except ValueError:
+            return False
+        return True
```
YukihiroSM, line 79: This one is good!

YukihiroSM (approved):
Nicely done!

PR #343, opened 2023-11-08
'Solution'

yul1valko:
Please let me know how I should use IntegerRange class in can_access function.


Polyakiv-Andrey (approved):
Good job!

PR #349, opened 2023-11-15
py-aquapark

ostboiko:


app/main.py
```diff
@@ -1,25 +1,71 @@
...
+from typing import Type, Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: "Visitor", oven: "Visitor") -> None:
```
EdAlekseiev: Use `Any` for `instance` and `oven`, because you don't know type of these arguments
```suggestion
    def __get__(self, instance: Any, oven: Any) -> None:
```

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: "Visitor", value: Any) -> None:
```
EdAlekseiev: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
v1

YukihiroSM (approved):
Good job!

PR #350, opened 2023-11-15
Solution

Marinel444:


app/main.py
```diff
@@ -1,25 +1,69 @@
...
+
+    def __set_name__(self, owner: Type[Any], name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: Any, objtype: Type[Any] = None) -> Any:
+        return getattr(obj, self.protected_name)
+
+    def __set__(self, obj: Any, value: int) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(obj, self.protected_name, value)
```
viktoria-rybenchuk, line 18: raise ValueError  when get incorrect value and add descriptive error message

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+    def validate(self) -> bool:
+        pass
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def validate(self) -> bool:
+        return (4 <= self.age <= 14
+                and 80 <= self.height <= 120
+                and 20 <= self.weight <= 50)
```
viktoria-rybenchuk: you should make validation by using `IntegerRange`

app/main.py
```diff
@@ -1,25 +1,69 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def validate(self) -> bool:
+        return (14 <= self.age <= 60
+                and 120 <= self.height <= 220
+                and 50 <= self.weight <= 120)
```
viktoria-rybenchuk: same here

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            if self.limitation_class(
+                age=visitor.age,
+                height=visitor.height,
+                weight=visitor.weight
+            ):
```
Dimosphen1, line 68: There is no need to create an if condition, you could just make an initialization of a `limitation_class`.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration. Besides that, consider placing arguments on a new line in case they have type annotations, it could improve the readability later on.

PR #358, opened 2023-11-22
'Solution_22112023'

Anton-Konyk:


app/main.py
```diff
@@ -2,24 +2,70 @@
...
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator
```
YukihiroSM: Wrong typehint. It should be Type[SlideLimitationValidator], because you are passing not Instance, but Type

YukihiroSM (changes requested):

YukihiroSM (approved):
Good job!

PR #361, opened 2023-12-01
Solution

pnasonov:


app/main.py
```diff
@@ -1,25 +1,76 @@
...
+            self,
+            instance: SlideLimitationValidator,
+            owner: SlideLimitationValidator
+    ) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, value)
+        else:
```
vasylhnatiuk: you can avoid else here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,25 +1,76 @@
+from __future__ import annotations
```
vsmutok, line 1: Remove  redundant  import

vsmutok (changes requested):

app/main.py
pnasonov: but flake8 and IDE give an error

app/main.py
```diff
@@ -1,25 +1,76 @@
...
+from typing import Type
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: SlideLimitationValidator, name: str) -> None:
```
ArsenPidhoretskyi: owner is not instance, owner - type

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #364, opened 2023-12-03
implemented "Aquapark"

380974261314kv:


app/main.py
```diff
@@ -1,25 +1,86 @@
...
+        )
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(
+            age=IntegerRange(4, 14),
+            height=IntegerRange(80, 120),
+            weight=IntegerRange(20, 50))
```
Dimosphen1: Do not use the `__init__` in the `ChildrenSlideLimitationValidator` and `AdultSlideLimitationValidator`, create class attributes with an assignment of `IntegerRange` with the corresponding values instead.

app/main.py
```diff
@@ -1,25 +1,86 @@
...
+
+    def validate(self, visitor: Visitor) -> bool:
+        age = visitor.age
+        height = visitor.height
+        weight = visitor.weight
+        return (
+            self.age.min_amount <= age <= self.age.max_amount
+            and self.height.min_amount <= height <= self.height.max_amount
+            and self.weight.min_amount <= weight <= self.weight.max_amount
+        )
```
Dimosphen1: There is no need for the `validate` method here, this logic is implemented in the `IntegerRange` class.

app/main.py
```diff
@@ -1,25 +1,86 @@
...
+        self,
+        name: str,
+        limitation_class: ChildrenSlideLimitationValidator
+            | AdultSlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.limitation_validator = limitation_class()
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_validator.validate(visitor)
```
Dimosphen1: Instead of the `validate` method, add a `try except` block where you will initialize a `limitation_validator` class instance with parameters from `visitor` and in case the `ValueError` is raised you should return `False` and `True` otherwise.

app/main.py
```diff
@@ -1,25 +1,86 @@
...
 class Slide:
-    pass
+    def __init__(
+        self,
+        name: str,
+        limitation_class: ChildrenSlideLimitationValidator
+            | AdultSlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.limitation_validator = limitation_class()
```
Dimosphen1: Perform `limitation_class` initialization in the `can_access` method to utilize the functionality of the `IntegerRange` class.

app/main.py
```diff
@@ -1,25 +1,86 @@
...
+            weight=IntegerRange(50, 120))
 
 
 class Slide:
-    pass
+    def __init__(
+        self,
+        name: str,
+        limitation_class: ChildrenSlideLimitationValidator
+            | AdultSlideLimitationValidator
```
Dimosphen1: Change your type annotation to indicate that you expect to obtain a subclass of `SlideLimitationValidator` class:
```suggestion
        limitation_class: type[SlideLimitationValidator]
```

app/main.py
```diff
@@ -1,25 +1,86 @@
...
+        return getattr(instance, self.name)
+
+    def __set__(self, instance: Any, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError(f"{self.name} must be between \
+                {self.min_amount} and {self.max_amount}")
+        setattr(instance, self.name, value)
+
+    def __set_name__(self, owner: Type, name: str) -> None:
+        self.name = name
```
Dimosphen1: Add an underscore before name to define the protected name:
```suggestion
        self.protected_name = "_" + name
```

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #365, opened 2023-12-05
Aqua-park

Ananiev-Vitalii:


app/main.py
```diff
@@ -1,25 +1,75 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
+
+    def __get__(self, instance: object, owner: type) -> Union:
+        return instance.__dict__[self.name]
```
EdAlekseiev, line 14: Don't use `__dict__` without exigency. You can use `getattr` here

app/main.py
```diff
@@ -1,25 +1,75 @@
...
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError(
+                f"{self.name} must be in the range"
+                f" {self.min_amount}-{self.max_amount}")
+        instance.__dict__[self.name] = value
```
EdAlekseiev, line 21: Use `setattr` here

app/main.py
```diff
@@ -1,25 +1,75 @@
...
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: tuple, weight: tuple, height: tuple) -> None:
+        self.age = IntegerRange(*age)
+        self.weight = IntegerRange(*weight)
+        self.height = IntegerRange(*height)
+
+    @abstractmethod
+    def validate(self, visitor: Visitor) -> bool:
+        pass
```
EdAlekseiev, line 40: This method is redundant

app/main.py
```diff
@@ -1,25 +1,75 @@
...
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: tuple, weight: tuple, height: tuple) -> None:
+        self.age = IntegerRange(*age)
+        self.weight = IntegerRange(*weight)
+        self.height = IntegerRange(*height)
```
EdAlekseiev, line 36: Variable of `IntegerRange` must be class attribute. Now they are instance attribute

app/main.py
```diff
@@ -1,25 +1,75 @@
...
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: tuple, weight: tuple, height: tuple) -> None:
```
EdAlekseiev, line 33: Invalid type hint
```suggestion
    def __init__(self, age: int, weight: int, height: int) -> None:
```

app/main.py
```diff
@@ -1,25 +1,75 @@
...
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
+        self.name = name
+        self.limitation_validator = limitation_class()
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_validator.validate(visitor)
```
EdAlekseiev, line 75: You shouldn't call method `validate`. All your validations must be in method `__set__` of IntegerRange

EdAlekseiev (changes requested):

app/main.py
Ananiev-Vitalii: @EdAlekseiev could you give me more hints? I don't quite understand what I need to do?

app/main.py
Ananiev-Vitalii: @EdAlekseiev could you give me more hints? I don't quite understand what I need to do?

PR #369, opened 2023-12-11
solve py-aquapark

zdmytryszyn:


app/main.py
```diff
@@ -1,25 +1,140 @@
...
+
+    def __set__(
+            self,
+            instance: SlideLimitationValidator,
+            value: int
+    ) -> None:
+        if instance.validate(value, self.min_amount, self.max_amount):
+            setattr(instance, self.private_name, value)
+        else:
+            setattr(instance, self.private_name, False)
```
Dimosphen1: Instead of delegating the logic of the validation to the `instance`, perform all required checks in the `__set__` method itself. You could raise an error and process it in the `can_access` method of the `Slide` class.

app/main.py
```diff
@@ -1,25 +1,140 @@
...
+            owner: SlideLimitationValidator,
+            name: str
+    ) -> None:
+        self.private_name = "_" + name
+        self.public_name = name
+
+    def __get__(
+            self,
+            instance: SlideLimitationValidator,
+            owner: Slide
```
Dimosphen1: Do not limit the use of the `IntegerRange` to only `SlideLimitationValidator` and `Slide` class instances. Use `Any` for `instance` and `object` for `owner` as an example. Change type annotations in all methods of `IntegerRange` where applicable.

app/main.py
```diff
@@ -1,25 +1,140 @@
...
+        self.height = height
+
+    @abstractmethod
+    def validate(
+            self,
+            value: int,
+            min_value: int,
+            max_value: int
+    ) -> bool:
+        return isinstance(value, int) and min_value <= value <= max_value
```
Dimosphen1: Do not declare and implement the `validate` method here and in the other classes, rely on the logic of the `__set__` method.

app/main.py
```diff
@@ -1,25 +1,140 @@
...
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(
+            self,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        super().__init__(age, weight, height)
```
Dimosphen1: There is no need to implement the `__init__` method here and in the `AdultSlideLimitationValidator` class, you could rely on the `__init__` of the parent class.

app/main.py
```diff
@@ -1,25 +1,140 @@
...
+    ) -> bool:
+        return super().validate(value, min_value, max_value)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Dimosphen1: You expect to obtain a subclass of the `SlideLimitationValidator` class, not an instance of the `SlideLimitationValidator` class. Change your type annotation:
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

app/main.py
```diff
@@ -1,25 +1,140 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, instance: Visitor) -> bool:
```
Dimosphen1, line 94: Change the logic to initialize the `limitation_class` in the try-except block and return the boolean value depending on the outcome.

Dimosphen1 (changes requested):

app/main.py
zdmytryszyn: changed the logic, used try-except block, but at this point of learning 'exception handling' module has not started yet, so what was the optimal way of resolving this issue without try-except  block?

app/main.py
zdmytryszyn: don't fully get why do we have to implement 'object' as a type annotations for 'owner'. is it because an instance of Any class is inherited from 'object' class?

Polyakiv-Andrey (approved):
Good job!

PR #370, opened 2023-12-11
Solution

KwintDimon:


app/main.py
```diff
@@ -2,24 +2,75 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: type  # SlideLimitationValidator
```
Dimosphen1: Remove the comment and indicate that you expect to obtain a subclass of `SlideLimitationValidator` using the following type annotation:
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

Dimosphen1 (changes requested):
Several minor changes were requested.

app/main.py
```diff
@@ -2,24 +2,75 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.public_name = name
```
viktoria-rybenchuk: where do you use public_name?

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> None:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            return
```
viktoria-rybenchuk: Raising a `ValueError`is a common practice in Python when a function or method receives an argument that is of the correct type but has an invalid value.

viktoria-rybenchuk: you can catch the `ValueError` in the `can_access `function using a `try-except `block.

app/main.py
```diff
@@ -2,24 +2,75 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> None:
```
viktoria-rybenchuk: Does this function return `None`

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
GJ!

PR #371, opened 2023-12-13
Solution

gertsovenko:


app/main.py
```diff
@@ -1,25 +1,108 @@
...
+
+    def __set__(
+            self,
+            instance: SlideLimitationValidator,
+            value: int
+    ) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_attr, value)
+        else:
+            setattr(instance, self.protected_attr, None)
```
viktoria-rybenchuk: Raising a `ValueError` is a common practice in Python when a function or method receives an argument that is of the correct type but has an invalid value.

app/main.py
```diff
@@ -1,25 +1,108 @@
...
+            weight: int,
+            height: int
+    ) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
+    def validate(self) -> bool:
+        pass
```
viktoria-rybenchuk: why do you need this method?

app/main.py
```diff
@@ -1,25 +1,108 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    weight = IntegerRange(20, 50)
+    height = IntegerRange(80, 120)
+
+    def validate(self) -> bool:
+        return True if self.age and self.weight and self.height else False
```

app/main.py
```diff
@@ -1,25 +1,108 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
+
+    def validate(self) -> bool:
+        return True if self.age and self.weight and self.height else False
```
viktoria-rybenchuk: this code looks redundant

app/main.py
```diff
@@ -1,25 +1,108 @@
...
+    ) -> bool:
+        if self.limitation_class == AdultSlideLimitationValidator:
+            person = AdultSlideLimitationValidator(
+                visitor.age, visitor.weight, visitor.height
+            )
+            return person.validate()
+        person = ChildrenSlideLimitationValidator(
+            visitor.age, visitor.weight, visitor.height
+        )
+        return person.validate()
```
viktoria-rybenchuk: you can catch the `ValueError`  using a try-except block. If it raised  `ValueError` then return False in other case True.

app/main.py
```diff
@@ -1,25 +1,108 @@
...
+            visitor: Visitor
+    ) -> bool:
+        if self.limitation_class == AdultSlideLimitationValidator:
+            person = AdultSlideLimitationValidator(
+                visitor.age, visitor.weight, visitor.height
+            )
+            return person.validate()
+        person = ChildrenSlideLimitationValidator(
+            visitor.age, visitor.weight, visitor.height
+        )
```
viktoria-rybenchuk: simplify the code by directly using `self.limitation_class` and passing the necessary values

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,25 +1,95 @@
...
+
+    def __set__(
+            self,
+            instance: SlideLimitationValidator,
+            value: int
+    ) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_attr, value)
+        else:
+            raise ValueError(f"Value {value} is not in range "
```
Polyakiv-Andrey: After opening the parentheses, write each the arguments on a new line, including the first one, the closing bracket is also written on a new line

Polyakiv-Andrey (changes requested):

app/main.py
gertsovenko: Do you mean it in raise here or everywhere in the code?

viktoria-rybenchuk (approved):
Good job!

PR #375, opened 2023-12-23
Solution

Grotski:


app/main.py
```diff
@@ -1,25 +1,100 @@
...
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> int:
+        return getattr(instance, self.name)
+
+    def __set__(self, instance: Any, value: int) -> None:
+        # if not self.min_amount <= value <= self.max_amount:
+        #     raise ValueError
+        setattr(instance, self.name, value)
```
```suggestion
        if not self.min_amount <= value <= self.max_amount:
             raise ValueError
        setattr(instance, self.name, value)
```

app/main.py
```diff
@@ -1,25 +1,100 @@
...
+            weight: IntegerRange,
+            height: IntegerRange
+    ) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
+    def validate(self, visitor: Visitor) -> bool:
+        pass
```

app/main.py
```diff
@@ -1,25 +1,100 @@
...
+            IntegerRange(20, 50),
+            IntegerRange(80, 120)
+        )
+
+    def validate(self, visitor: Visitor) -> bool:
+        return all([
+            self.age.min_amount <= visitor.age <= self.age.max_amount,
+            self.height.min_amount <= visitor.height <= self.height.max_amount,
+            self.weight.min_amount <= visitor.weight <= self.weight.max_amount
+        ])
```
EdAlekseiev: Delete this method, please

app/main.py
```diff
@@ -1,25 +1,100 @@
...
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_class.validate(visitor)
+        # try:
+        #     self.limitation_class.age = visitor.age
+        #     self.limitation_class.weight = visitor.weight
+        #     self.limitation_class.height = visitor.height
+        #     return True
+        # except ValueError:
+        #     return False
```
```suggestion
        try:
            self.limitation_class.age = visitor.age
            self.limitation_class.weight = visitor.weight
            self.limitation_class.height = visitor.height
            return True
        except ValueError:
            return False
```

EdAlekseiev (changes requested):

app/main.py
Grotski: I can't get the code to work with this logic

EdAlekseiev (changes requested):
Fix flake8

Grotski:
I can't get the code to work with this logic

        try:
            self.limitation_class.age = visitor.age
            self.limitation_class.weight = visitor.weight
            self.limitation_class.height = visitor.height
            return True
        except ValueError:
            return False

EdAlekseiev (approved):
Good job

Grotski:
but why good job? i pass only 6 tests

PR #386, opened 2024-01-04
Solution

oboliziuk:


app/main.py
```diff
@@ -1,25 +1,73 @@
...
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.private_name = "_" + name
+
+    def __get__(self, obj: Any, objtype: Any = None) -> int:
+        return getattr(obj, self.private_name)
+
+    def __set__(self, obj: Any, value: int) -> None:
+        setattr(obj, self.private_name, value)
```
ArsenPidhoretskyi, line 21: You should validate that value in range of min and max

app/main.py
```diff
@@ -1,25 +1,73 @@
...
+    def __init__(self, age: tuple[int, int],
+                 weight: tuple[int, int],
+                 height: tuple[int, int]) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
+    def validate(self, age: int, height: int, weight: int) -> bool:
+        pass
```
ArsenPidhoretskyi: This method is redundant
Fix in all places!

ArsenPidhoretskyi (changes requested):
You totally does not use descriptor

app/main.py
```diff
@@ -1,25 +1,73 @@
...
-    pass
+    def __init__(self) -> None:
+        super().__init__(age=(4, 14), weight=(20, 50), height=(80, 120))
+
+    def validate(self, age: int, height: int, weight: int) -> bool:
+        age_valid = self.age[0] <= age <= self.age[1]
+        height_valid = self.height[0] <= height <= self.height[1]
+        weight_valid = self.weight[0] <= weight <= self.weight[1]
+
+        return age_valid and height_valid and weight_valid
```
ArsenPidhoretskyi: This method is redundant

app/main.py
```diff
@@ -1,25 +1,73 @@
...
+                 weight: tuple[int, int],
+                 height: tuple[int, int]) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
```
ArsenPidhoretskyi: Why do you not use IntegerRange as class property for descriptors

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Looks Good!

PR #391, opened 2024-01-11
Solution

Damodar-kumar:


app/main.py
```diff
@@ -2,24 +2,64 @@
...
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: type, owner: type) -> int | str:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: type, value: int) -> None:
+        if value in range(self.min_amount, self.max_amount + 1):
+            setattr(instance, self.protected_name, True)
+        else:
+            setattr(instance, self.protected_name, False)
```
ArsenPidhoretskyi: No in case not in rage raise error, if in you should set value

app/main.py
```diff
@@ -2,24 +2,64 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: it accepts type, not instance

app/main.py
```diff
@@ -2,24 +2,64 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+        self.final = True
```
ArsenPidhoretskyi: this is redundant

app/main.py
```diff
@@ -2,24 +2,64 @@
...
+
+    def can_access(self, instance: Visitor) -> bool:
+        final = self.limitation_class(
+            instance.age,
+            instance.weight,
+            instance.height
+        )
+        return final.age and final.height and final.weight is True
```
ArsenPidhoretskyi: use try-excpet

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #398, opened 2024-01-16
Solution

eLQeR:


app/main.py
```diff
@@ -1,25 +1,66 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
YukihiroSM: In this case you are providing a class as limitation class, so the typehint would be: 
```suggestion
            limitation_class: type(SlideLimitationValidator)
```

YukihiroSM (changes requested):
Else looks good to me!

NikOneZ1 (approved):

PR #404, opened 2024-01-17
Solution

Meksim-fesh:


app/main.py
```diff
@@ -1,25 +1,76 @@
...
+        self.protected_name = f"_{name}"
+
+    def __get__(self, instance: IntegerRange, owner: IntegerRange) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: IntegerRange, value: int) -> None:
+        if (self.min_amount <= value <= self.max_amount):
+            setattr(instance, self.protected_name, True)
+        else:
+            setattr(instance, self.protected_name, False)
```
ArsenPidhoretskyi: No, in this case you will not know actual value, in case not in range you should raise an exception, otherwise set correct value

app/main.py
```diff
@@ -1,25 +1,76 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: limitation_class is type of SlideLimitationValidator

ArsenPidhoretskyi (changes requested):

app/main.py
vasylhnatiuk: fix it

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #419, opened 2024-02-02
Solution

MykhailoIvchenko:


app/main.py
```diff
@@ -1,25 +1,78 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    weight = IntegerRange(20, 50)
+    height = IntegerRange(80, 120)
+
+    def __init__(self, age: int, height: int, weight: int) -> None:
+        super().__init__(age, height, weight)
```

app/main.py
```diff
@@ -1,25 +1,78 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, height: int, weight: int) -> None:
+        super().__init__(age, height, weight)
```

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+
+    def __get__(self, instance: Any, objtype: Any = None) -> None:
+        value = getattr(instance, self.protected_name)
+        return value
+
+    def __set__(self, instance: Any, value: Any) -> None:
+        if self.validate(value):
+            setattr(instance, self.protected_name, value)
+        else:
+            setattr(instance, self.protected_name, None)
```
EdAlekseiev: You should raise ValueError in this case

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+
+    def can_access(self, visitor: Visitor) -> bool:
+        key_params = self.limitation_class(visitor.age, visitor.height,
+                                           visitor.weight)
+
+        for attr_name, attr_value in vars(key_params).items():
+            if attr_value is None:
+                return False
+
+        return True
```
EdAlekseiev: Use try...except here

EdAlekseiev (changes requested):

app/main.py
MykhailoIvchenko: Fixed

app/main.py
MykhailoIvchenko: Fixed

app/main.py
MykhailoIvchenko: Fixed

app/main.py
MykhailoIvchenko: Fixed

EdAlekseiev (approved):
Good job

PR #421, opened 2024-02-04
solution

MilArtem78:


app/main.py
```diff
@@ -2,24 +2,84 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: It accepts type of SlideLimitationValidator, not instance

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,24 +2,84 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: type(SlideLimitationValidator)
```
Dimosphen1, line 75: In case of type annotations, you could indicate that you expect to obtain a subclass of `SlideLimitationValidator` class using the following syntax
```suggestion
            limitation_class: type[SlideLimitationValidator]
```
Use this syntaxis in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #425, opened 2024-02-07
solution

Eugen-Veklich:


app/main.py
```diff
@@ -1,25 +1,74 @@
...
+                and self.height.min_value <= visitor.height
+                <= self.height.max_value
+                and self.weight.min_value <= visitor.weight
+                <= self.weight.max_value)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator) -> None:
```
YukihiroSM:
```suggestion
                 limitation_class: type(SlideLimitationValidator)) -> None:
```

As you are providing a class as a parameter - typehint should be like this.

app/main.py
```diff
@@ -1,25 +1,74 @@
...
+        super().__init__(age=IntegerRange(4, 14),
+                         height=IntegerRange(80, 120),
+                         weight=IntegerRange(20, 50))
+
+    def validate(self, visitor: Visitor) -> bool:
+        return (self.age.min_value <= visitor.age <= self.age.max_value
+                and self.height.min_value <= visitor.height
+                <= self.height.max_value
+                and self.weight.min_value <= visitor.weight
+                <= self.weight.max_value)
```
YukihiroSM: You don't need to have init and validate methids here as you implemented logic in IntegerRange. 

This class should only contain something like: 
    age = IntegerRange(4, 14)
    weight = IntegerRange(20, 50)

app/main.py
```diff
@@ -1,25 +1,74 @@
...
+        super().__init__(age=IntegerRange(14, 60),
+                         height=IntegerRange(120, 220),
+                         weight=IntegerRange(50, 120))
+
+    def validate(self, visitor: Visitor) -> bool:
+        return (self.age.min_value <= visitor.age <= self.age.max_value
+                and self.height.min_value <= visitor.height
+                <= self.height.max_value
+                and self.weight.min_value <= visitor.weight
+                <= self.weight.max_value)
```
YukihiroSM: same comment as for ChildrenSlideLimitationValidator

app/main.py
```diff
@@ -1,25 +1,74 @@
...
+        self.max_value = max_value
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
+
+    def __get__(self, instance: object, owner: type) -> tuple[int, int]:
+        return self.min_value, self.max_value
+
+    def __set__(self, instance: object, value: str) -> None:
+        self.min_value, self.max_value = value
```
YukihiroSM: You should validate it here and raise ValueError in wrong case.

app/main.py
```diff
@@ -1,25 +1,74 @@
...
-    pass
+    def __init__(self, age: IntegerRange,
+                 weight: IntegerRange, height: IntegerRange) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
+    def validate(self, visitor: Visitor) -> None:
+        pass
```
YukihiroSM: This can be removed after you will refactor child classes

app/main.py
```diff
@@ -1,25 +1,74 @@
...
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator) -> None:
+        self.name = name
+        self.limitation_class = limitation_class()
+
+    def can_access(self, visitor: Visitor) -> None:
+        return self.limitation_class.validate(visitor)
```
YukihiroSM: try to create an instance of limitation_class and except ValueError

YukihiroSM (changes requested):

YukihiroSM (approved):
Nicely done! Good job on refactoring this one!

PR #430, opened 2024-02-11
Solution

serji13:


app/main.py
```diff
@@ -1,25 +1,74 @@
...
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any) -> None:
+        setattr(instance, self.protected_name)
```
EdAlekseiev: You should make validation and raise ValueError inside this merhod

EdAlekseiev (changes requested):

app/main.py
serji13: This task is very complicated. I generally think that there are errors in the condition and in the tests. In my solution, the setter is not used at all. I was hoping that you would make more in-depth comments.

app/main.py
```diff
@@ -1,25 +1,81 @@
...
+    def __init__(self) -> None:
+        pass
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        self.age = IntegerRange(4, 14)
+        self.height = IntegerRange(80, 120)
+        self.weight = IntegerRange(20, 50)
```
```suggestion
    age = IntegerRange(4, 14)
    height = IntegerRange(80, 120)
    weight = IntegerRange(20, 50)
```

app/main.py
```diff
@@ -1,25 +1,81 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        self.age = IntegerRange(14, 60)
+        self.height = IntegerRange(120, 220)
+        self.weight = IntegerRange(50, 120)
```
```suggestion
    def __init__(self) -> None:
        age = IntegerRange(14, 60)
        height = IntegerRange(120, 220)
        weight = IntegerRange(50, 120)
```

app/main.py
```diff
@@ -1,25 +1,81 @@
...
+                )
+                and (
+                    self.limitation_class.weight.min_amount <= visitor.weight
+                    <= self.limitation_class.weight.max_amount
+                )
+                and (
+                    self.limitation_class.height.min_amount <= visitor.height
+                    <= self.limitation_class.height.max_amount
+                )
+        ):
```
EdAlekseiev: You should create instance of `limitation_class` and catch `ValueErorr`

EdAlekseiev (changes requested):

app/main.py
serji13: I really want to deal with this task. But I can't. Could you send me a complete working solution that passes all checks - so that I can understand everything with an example.

app/main.py
serji13: If I remove "self .", in the "__init __" method, this code just stops working.

EdAlekseiev (changes requested):
If you stuck with this task, you can write to chat of your group or visit QnA

serji13:
Now I create an instance of the Slide class without using IntegerRange, and only when the can_access function is launched - IntegerRange is connected. My main mistake was that I believed that an instance of the Slide class should be created in advance using the IntegerRange function. And can_access should only check whether the existing Visitor can have access to the existing Slide, which already has restrictions. Based on my previous assumption, filling half the classes did not make sense. :-(

EdAlekseiev (approved):

PR #432, opened 2024-02-12
Solution

kolya2731:


app/main.py
```diff
@@ -1,25 +1,90 @@
...
+            max_amount
+        )
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Dimosphen1, line 83: You should indicate in the type annotation that you expect to obtain a subclass of `SlideLimitationValidator`, not an instance of this class. Change your type annotation to encounter that.

app/main.py
```diff
@@ -1,25 +1,90 @@
...
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
+    def is_valid(self, visitor: Visitor) -> None:
```
Dimosphen1, line 39: There is no need for `is_valid` methods in all classes, you should define `age`, `weight` and `height` as class attributes and assign corresponding instances of `IntegerRange` class with limitations to each range. It will allow to utilize the `__set__` method later to trigger an error for further processing.

app/main.py
```diff
@@ -1,25 +1,90 @@
...
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
+    ) -> None:
+
+        self.name = name
+        self.limitation_validator = limitation_class()
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_validator.is_valid(visitor)
```
Dimosphen1, line 90: You should try initializing the `limitation_class` with values present in `visitor` and depending on the presence or absence of an exception (`ValueError`) return the boolean value.

Dimosphen1 (changes requested):

PR #436, opened 2024-02-16
Solution

vburkalo:


app/main.py
```diff
@@ -1,25 +1,67 @@
...
+        self.height = height
+        self.age = age
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    @staticmethod
+    @abstractmethod
+    def validate(visitor: Visitor) -> bool:
+        pass
```
EdAlekseiev: There is no need for `valid` methods in all classes, you should define `age`, `weight` and `height` as class attributes and assign instances of IntegerRange class with limitations to each range. It will allow to utilize the `__set__` method later to trigger an error for further processing.

app/main.py
```diff
@@ -1,25 +1,67 @@
...
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator
+                 ) -> None:
+        self.name = name
+        self.limitation_validator = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_validator.validate(visitor)
```
EdAlekseiev: You should try initializing the `limitation_class` with values from `visitor`.

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -2,24 +2,69 @@
...
+        self.height = height
+        self.age = age
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self,
+                 age: IntegerRange,
+                 weight: IntegerRange,
+                 height: IntegerRange
```
```suggestion
                 age: int,
                 weight: int,
                 height: int
```

EdAlekseiev (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #440, opened 2024-02-25
Solution

MariiaSydoruk:


app/main.py
```diff
@@ -1,25 +1,65 @@
-from abc import ABC
+class IntegerRange:
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
 
+    def __set_name__(
+            self, owner: "SlideLimitationValidator",
+            name: str) -> None:
```
Dimosphen1: If there are no arguments on the same line as the opening parenthesis, the closing parenthesis should not have any arguments on the same line as well. Apply this change to all related places

Dimosphen1: Place each argument on a new line including `self`

app/main.py
```diff
@@ -1,25 +1,65 @@
-from abc import ABC
+class IntegerRange:
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
 
+    def __set_name__(
+            self, owner: "SlideLimitationValidator",
```
Dimosphen1: Do not limit `owner` to just instances of `SlideLimitationValidator`, indicate that it could be any object in the type annotation to not limit the `IntegerRange` class. Apply this change to all related places

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+            self, name: str,
+            limitation_class: SlideLimitationValidator) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+            return True
+        except ValueError:
```
Dimosphen1: You should catch here not only `ValueError`, but `TypeError` as well since it is declared

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+            self, name: str,
+            limitation_class: SlideLimitationValidator) -> None:
```
Dimosphen1: Indicate that you expect to obtain not an instance of the `SlideLimitationValidator` class, but a subclass of the `SlideLimitationValidator` class for `limitation_class` in the type annotation

app/main.py
```diff
@@ -1,25 +1,65 @@
...
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
-class SlideLimitationValidator(ABC):
-    pass
+class SlideLimitationValidator:
```
Dimosphen1: Keep `SlideLimitationValidator` class inheritance from the `ABC`, do not remove it

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -2,24 +2,70 @@
...
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: (
+                ChildrenSlideLimitationValidator
+                | AdultSlideLimitationValidator
+            )
```
Dimosphen1: Use the following syntax for the type annotation here
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

app/main.py
```diff
@@ -2,24 +2,70 @@
...
+            self,
+            owner: any,
+            name: str
+    ) -> None:
+        self.protected_name = f"_{name}"
+
+    def __get__(self, instance: any, owner: any) -> str:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: any, value: int) -> None:
```
Dimosphen1: It is better to use `Any` from `typing` since it is not deprecated. Change in all places

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #450, opened 2024-03-07
Solution

liudmylaperchevska:


app/main.py
```diff
@@ -1,25 +1,82 @@
...
+            <= slide.age.max_amount,
+
+            slide.weight.min_amount
+            <= visitor.weight
+            <= slide.weight.max_amount,
+
+            slide.height.min_amount
+            <= visitor.height
+            <= slide.height.max_amount
+        ])
```
Dimosphen1: There is a code duplication, please utilize the `__set__` method from the `IntegerRange` class and raise an exception that would be processed here and thus allow avoiding code duplication.

app/main.py
```diff
@@ -1,25 +1,82 @@
...
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class:
+            AdultSlideLimitationValidator | ChildrenSlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class()
```
Dimosphen1: You should call `limitation_class` in the `can_access` method to trigger an exception raise

app/main.py
```diff
@@ -1,25 +1,82 @@
...
+        )
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class:
+            AdultSlideLimitationValidator | ChildrenSlideLimitationValidator
```
Dimosphen1: Indicate that `limitation_class` should be a subclass of `SlideLimitationValidator`, not an instance of the mentioned classes
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

app/main.py
```diff
@@ -1,25 +1,82 @@
...
+        self.height = height
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(
+            age=IntegerRange(4, 14),
+            weight=IntegerRange(20, 50),
+            height=IntegerRange(80, 120)
```
Dimosphen1: You should define these arguments as class attributes to utilize the functionality of the `IntegerRange` class here and in the `AdultSlideLimitationValidator` class

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,25 +1,81 @@
...
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    #
+    # age = IntegerRange
+    # weight = IntegerRange
+    # height = IntegerRange
```
Dimosphen1: Delete unnecessary comments

app/main.py
```diff
@@ -1,25 +1,81 @@
...
+            type[SlideLimitationValidator]
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+            return True
+        except ValueError:
```
Dimosphen1: You should catch `TypeError` here as well since it could be raised in the `__set__` method

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,25 +1,77 @@
...
+            type[SlideLimitationValidator]
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+            return True
+        except ValueError or TypeError:
```
Dimosphen1, line 76: It is preferable to catch multiple errors in this way
```suggestion
        except (ValueError, TypeError):
```

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #453, opened 2024-03-11
Solution

jyjuk:


app/main.py
```diff
@@ -2,24 +2,64 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Dimosphen1: You should indicate that you expect to obtain not an instance of the `SlideLimitationValidator` class for `limitation_class` but rather its subclass. Use the following type annotation for that
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

Dimosphen1 (changes requested):
Several minor changes were requested

vasylhnatiuk (approved):

PR #456, opened 2024-03-11
implement the Aqua-park validate logic

ghost:


app/main.py
```diff
@@ -2,24 +2,77 @@
...
+        self.private_attribute = f"_{name}"
+
+    def __get__(self, instance: object, owner: object) -> bool:
+        return getattr(instance, self.private_attribute)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if self.max_amount >= value >= self.min_amount:
+            setattr(instance, self.private_attribute, True)
+        else:
+            setattr(instance, self.private_attribute, False)
```
ArsenPidhoretskyi: You should set original value, in other case raise an exception

ArsenPidhoretskyi (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):
you pushed nothing

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+
+    def __init__(self,
+                 name: str,
+                 limitation_class: type(SlideLimitationValidator)
```
```suggestion
                 limitation_class: type[SlideLimitationValidator]
```

ArsenPidhoretskyi (changes requested):
No need to call type() use type[] for annotations

vsmutok (approved):
GJ!

PR #465, opened 2024-03-17
'Solution'

donbas777:


app/main.py
```diff
@@ -1,25 +1,82 @@
...
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: "IntegerRange",
+                 weight: "IntegerRange",
+                 height: "IntegerRange"
+                 ) -> None:
```
Dimosphen1: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Add this change to all classes where relevant

app/main.py
```diff
@@ -1,25 +1,82 @@
...
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
+    def validate(self, visitor: "Visitor") -> bool:
```
Dimosphen1: Delete this method from all classes, utilize the descriptor class functionality instead

app/main.py
```diff
@@ -1,25 +1,82 @@
...
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: "IntegerRange",
```
Dimosphen1: Indicate here that you expect to obtain integer instead of an instance of `IntegerRange` class

app/main.py
```diff
@@ -1,25 +1,82 @@
...
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: "SlideLimitationValidator"
+                 ) -> None:
+        self.name = name
+        self.limitation_validator = limitation_class()
+
+    def can_access(self, visitor: "Visitor") -> bool:
+        return self.limitation_validator.validate(visitor)
```
Dimosphen1: Create a try-except block and initialize limitation class inside of it to utilize the `__set__` functionality

app/main.py
```diff
@@ -1,25 +1,82 @@
...
+            self.height.min_amount <= visitor.height <= self.height.max_amount
+        ])
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(IntegerRange(14, 60),
+                         IntegerRange(50, 120),
+                         IntegerRange(120, 220))
```
Dimosphen1: Define class attributes using `IntegerRange` class with corresponding values in the subclasses of the `SlideLimitationValidator` class

app/main.py
```diff
@@ -1,25 +1,82 @@
...
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: "SlideLimitationValidator"
+                 ) -> None:
+        self.name = name
+        self.limitation_validator = limitation_class()
+
+    def can_access(self, visitor: "Visitor") -> bool:
```
Dimosphen1: Do not use classes for annotations in double quotes, omit them
```suggestion
    def can_access(self, visitor: Visitor) -> bool:
```
In case you face errors, add `from __future__ import annotations`

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -2,24 +2,97 @@
...
+            IntegerRange(20, 50),
+            IntegerRange(80, 120)
+        )
+
+    def validate(self, visitor: Visitor) -> bool:
+        return all([
+            self.age.min_amount <= visitor.age <= self.age.max_amount,
+            self.weight.min_amount <= visitor.weight <= self.weight.max_amount,
+            self.height.min_amount <= visitor.height <= self.height.max_amount
+        ])
```
viktoria-rybenchuk: there is no need to redefine validate method, you have already validated this value in IntegerRange

app/main.py
```diff
@@ -2,24 +2,97 @@
...
+            IntegerRange(50, 120),
+            IntegerRange(120, 220)
+        )
+
+    def validate(self, visitor: Visitor) -> bool:
+        return all([
+            self.age.min_amount <= visitor.age <= self.age.max_amount,
+            self.weight.min_amount <= visitor.weight <= self.weight.max_amount,
+            self.height.min_amount <= visitor.height <= self.height.max_amount
+        ])
```

app/main.py
```diff
@@ -2,24 +2,97 @@
...
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.limitation_validator = None
+        try:
+            self.limitation_validator = limitation_class()
+        except Exception as e:
+            print(f"Error initializing limitation validator: {e}")
```
viktoria-rybenchuk: there is no need for using try-except in init method

app/main.py
```diff
@@ -2,24 +2,97 @@
...
+
+    def can_access(self, visitor: Visitor) -> bool:
+        if self.limitation_validator:
+            try:
+                return self.limitation_validator.validate(visitor)
```
viktoria-rybenchuk: you should pass in limitation class all needed arguments from visitor

viktoria-rybenchuk (changes requested):

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
```diff
@@ -2,24 +2,91 @@
...
-    pass
+    def __init__(
+            self,
+            age: IntegerRange(int, int),
+            weight: IntegerRange(int, int),
+            height: IntegerRange(int, int)
+    ) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
```
```suggestion
    def __init__(
            self,
            age: int,
            weight: int,
            height: int
    ) -> None:
        self.age = age
        self.weight = weight
        self.height = height
```

app/main.py
```diff
@@ -2,24 +2,91 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(
+            IntegerRange(4, 14),
+            IntegerRange(20, 50),
+            IntegerRange(80, 120)
+        )
```
```suggestion
        age = IntegerRange(4, 14)
        weight = IntegerRange(20, 50)
        height  = IntegerRange(80, 120)
```
Set them as descriptors

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
```diff
@@ -2,24 +2,91 @@
...
+            self.height.min_amount <= visitor.height <= self.height.max_amount
+        ])
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: it accepts type of SlideLimitationValidator

ArsenPidhoretskyi (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #470, opened 2024-03-22
Solution

Line181:


app/main.py
```diff
@@ -1,25 +1,84 @@
...
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(
+                age=visitor.age,
+                weight=visitor.weight,
+                height=visitor.height
+            )
+        except ValueError:
+            return False
+        else:
```
Dimosphen1: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

app/main.py
```diff
@@ -1,25 +1,84 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Dimosphen1: Indicate in the type annotation that you expect to obtain not an instance of the `SlideLimitationValidator` class, but rather its subclass
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

Dimosphen1 (changes requested):
Several minor changes were requested

vasylhnatiuk (approved):

PR #473, opened 2024-03-24
hard one

panicua:
I gotta admit that I stucked at this one for a long time, frist of all I didn't know how to handle ```raise``` cause we didn't learn ```try/except``` yet and I decided just to set the attribute to ```False``` so it's easy to handle later and no error occurs in tests. Of course I could use ```try/except``` but I thought it's not how it ment to be written. If you have a suggestion how we can handle it better without using ```try/except``` I would listen to it, or if you want me to rewrite it to ```try/except``` with ```raise``` just tell.
NVIDIA_Share_aVRnnTsdHN

And this isn't even stopping here, I don't know if I put right type annotations cause with these ```get/set``` methods it's getting harder to keep track of annotations.

And one of the hardest parts that I had "cheat" (check someones else pull request) cause we aren't inheriting ```Visitor``` class to ```ChildrenSlideLimitationValidator / AdultSlideLimitationValidator``` and not even their parent class ```SlideLimitationValidator(ABC)``` inherit it. I didn't know how to fill them right I was trying to do something like this:
NVIDIA_Share_L5IavdyNQd

But it wasn't right cause we aren't inheriting ```Visitor``` class and I didn't know what's wrong with my code and what is the other way to pass that ```Visit``` class data, so I looked at one of students pull request to figure it out, it appeared that it to be like this, it made me feel bad that I didn't see it:
NVIDIA_Share_Hoj80pAsAs

Overall it feels like I had lots of pain with this task and I'll have to rewrite some of it cause it doesn't seem really good ;(

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -16,8 +16,12 @@ class IntegerRange:
     def __set__(self,
                 instance: type[SlideLimitationValidator],
                 value: int) -> None:
+        if not isinstance(value, (int, float)):
+            raise TypeError("Value can only be int or float")
         if not self.min_amount <= value <= self.max_amount:
-            return setattr(instance, self.protected_name, False)
+            raise ValueError(f"The value of *{self.protected_name[1:]}* "
+                             f"should be in range({self.min_amount} - "
+                             f"{self.max_amount}). Yours ({value})")
         setattr(instance, self.protected_name, value)
 
     def __set_name__(self,
@@ -68,8 +72,12 @@ class Slide:
         self.limitation_class = limitation_class
 
     def can_access(self, visitor: Visitor) -> bool:
-        lim = self.limitation_class(age=visitor.age,
-                                    height=visitor.height,
-                                    weight=visitor.weight)
-
-        return (lim.age and lim.weight and lim.height) is not False
+        try:
+            self.limitation_class(age=visitor.age,
+                                  height=visitor.height,
+                                  weight=visitor.weight)
+        except ValueError as e:
+            print(e)
+            return False
+        else:
+            return True
--- a/app/main.py
+++ b/app/main.py
@@ -76,7 +76,6 @@ class Slide:
             self.limitation_class(age=visitor.age,
                                   height=visitor.height,
                                   weight=visitor.weight)
-        except ValueError as e:
-            print(e)
+        except ValueError:
             return False
         return True
--- a/app/main.py
+++ b/app/main.py
@@ -79,5 +79,4 @@ class Slide:
         except ValueError as e:
             print(e)
             return False
-        else:
-            return True
+        return True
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,25 +1,75 @@
...
+    def __get__(self,
+                instance: type[SlideLimitationValidator],
+                owner: SlideLimitationValidator) -> int | None:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self,
+                instance: type[SlideLimitationValidator],
+                value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            return setattr(instance, self.protected_name, False)
```
oleksandra-shershen: In my opinion, it is better to use:
```
if not (self.min_value <= value <= self.max_value):
            raise ValueError(
                f"Value must be between {self.min_value} and {self.max_value}"
            )
```

app/main.py
```diff
@@ -1,25 +1,75 @@
...
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: type[SlideLimitationValidator]) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
```
oleksandra-shershen, line 74: in this method you can try to use try-catch block, like this:
```
        try:
            return True
            return False
```

oleksandra-shershen (changes requested):
Check my comments
GJ!

oleksandra-shershen (approved):
GJ!

app/main.py
```diff
@@ -1,25 +1,83 @@
...
+                instance: type[SlideLimitationValidator],
+                owner: SlideLimitationValidator) -> int | None:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self,
+                instance: type[SlideLimitationValidator],
+                value: int) -> None:
+        if not isinstance(value, (int, float)):
+            raise TypeError("Value can only be int or float")
+        if not self.min_amount <= value <= self.max_amount:
```
oleksandra-shershen: add `elif not` instead of `if not`

oleksandra-shershen (approved):
Change if not to elif not

oleksandra-shershen (approved):
GJ 2.0!

app/main.py
```diff
@@ -1,25 +1,82 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(age=visitor.age,
+                                  height=visitor.height,
+                                  weight=visitor.weight)
+        except ValueError as e:
+            print(e)
```
YuriiKindrat: remove unnecessary pring here )

YuriiKindrat (approved):
LGTM!

PR #474, opened 2024-03-25
Solution for task aquapark

oleksandra-shershen:


Boyok1337 (approved):
GJ!

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+                f"Value must be between {self.min_value} and {self.max_value}"
+            )
+        setattr(obj, self.protected_name, value)
 
 
 class Visitor:
-   pass
+    age = IntegerRange(0, 150)
+    weight = IntegerRange(0, 500)
+    height = IntegerRange(0, 300)
```
vitalinamal: I don`t think we need this. It can work without this

vitalinamal (approved):
Good job!

app/main.py
YuriiKindrat: some really old guy will be really sad (

YuriiKindrat (approved):
GJ!

app/main.py
oleksandra-shershen: Resolved

PR #477, opened 2024-03-26
Solution

DmytroHlazyrin:


app/main.py
```diff
@@ -1,25 +1,97 @@
...
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError("Value must be an integer")
+        if not (self.min_amount <= value <= self.max_amount):
```
oleksandra-shershen: you can use `elif` instead of `if`

app/main.py
```diff
@@ -1,25 +1,97 @@
...
+            raise ValueError(
+                f"Value must be between {self.min_amount} "
+                f"and {self.max_amount}"
+            )
+        setattr(instance, self.protected_name, value)
 
 
 class Visitor:
-   pass
+    age = IntegerRange(0, 120)
```
oleksandra-shershen: We don't need to create instances of IntegerRange class in Visitor class. So you can remove these lines of code:
```
age = IntegerRange(0, 120)
weight = IntegerRange(0, 300)
height = IntegerRange(20, 300)
```

app/main.py
```diff
@@ -1,25 +1,97 @@
...
+
+    @abstractmethod
+    def can_use_slide(self, visitor: Visitor) -> Any:
+        pass
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(IntegerRange(4, 14),
```
oleksandra-shershen: In ChildrenSlideLimitationValidator you can remove `__init__()`

app/main.py
```diff
@@ -1,25 +1,97 @@
...
+            and self.height.min_amount
+            <= visitor.height <= self.height.max_amount
+            and self.weight.min_amount
+            <= visitor.weight <= self.weight.max_amount
+        )
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
```
oleksandra-shershen: Remove `__init__()`

app/main.py
```diff
@@ -1,25 +1,97 @@
...
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(IntegerRange(4, 14),
+                         IntegerRange(80, 120),
+                         IntegerRange(20, 50))
+
+    def can_use_slide(self, visitor: Visitor) -> bool:
+        return (
+            self.age.min_amount <= visitor.age <= self.age.max_amount
```
oleksandra-shershen: Can you rewrite the code to make it look like this? Because this code is hard to read
```
    self.age.min_amount <= visitor.age <= self.age.max_amount
    and self.height.min_amount <= visitor.height <= self.height.max_amount
    and self.weight.min_amount <= visitor.weight <= self.weight.max_amount
```

Or second variant:
````
    age_condition = self.age.min_amount <= visitor.age <= self.age.max_amount
    height_condition = self.height.min_amount <= visitor.height <= self.height.max_amount
    weight_condition = self.weight.min_amount <= visitor.weight <= self.weight.max_amount

    return all([age_condition, height_condition, weight_condition])

File: app/main.py [hunk 6]
```diff
@@ -1,25 +1,97 @@
...
+            and self.weight.min_amount
+            <= visitor.weight <= self.weight.max_amount
+        )
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: [SlideLimitationValidator]) -> None:
```
oleksandra-shershen: Why we need [] here?

oleksandra-shershen (changes requested):
Check comments)

app/main.py
DmytroHlazyrin: Yes, sure

app/main.py
DmytroHlazyrin: I follow a slightly different logic, which means using instances of the SlideLimitationValidator class. So I need __init__

app/main.py
DmytroHlazyrin: If you look closely, you'll see that my version is exactly like your first one. I moved the line because it was too long.

oleksandra-shershen (approved):
GJ!

app/main.py
```diff
@@ -2,24 +2,77 @@
...
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: IntegerRange,
+                 height: IntegerRange,
+                 weight: IntegerRange
+                 ) -> None:
```
YuriiKindrat: bad formatting here (

app/main.py
```diff
@@ -2,24 +2,77 @@
...
+            <= visitor.weight <= self.weight.max_amount
+        )
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(IntegerRange(4, 14),
+                         IntegerRange(80, 120),
+                         IntegerRange(20, 50))
```
YuriiKindrat: no need to use init here)

app/main.py
YuriiKindrat: @oleksandra-shershen is right

YuriiKindrat (changes requested):

Oleksl888 (approved):
Well done!

Oleksl888 (approved):

PR #479, opened 2024-03-26
Solution

CharmNiovi:


app/main.py
```diff
@@ -2,24 +2,68 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
CharmNiovi, line 59: Whether it is a mistake to do so:
`limitation_class: SlideLimitationValidator.__subclasses__`
Because in this case, the PyCharm emphasizes `self.limitation_class(visitor.age, visitor.weight, visitor.height)` with the words "'SlideLimitationValidator' object is not callable"

Oleksl888: Pycharm thinks you are passing it an instance of a class not the class - you can use `Type[SlideLimitationValidator]` to specify that you passing class itself as a parameter

Oleksl888 (approved):

PR #480, opened 2024-03-27
Solution

rzaliznyi:


app/main.py
```diff
@@ -2,24 +2,83 @@
...
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.public_name = name
+        self.private_name = "_" + name
+
+    def __get__(self, obj: object, objtype: None) -> int:
+        value = getattr(obj, self.private_name)
+        return value
+
+    def __set__(self, obj: object, value: int) -> None:
+        setattr(obj, self.private_name, value)
```
vsmutok, line 25:
```suggestion
        if not isinstance(value, int):
            raise ValueError("Value must be integer.")
        elif not (self.min_amount <= value <= self.max_amount):
            raise ValueError(
                f"Value must be between {self.min_amount}"
                f" and {self.max_amount}"
            )
        setattr(instance, self.protected_name, value)
```

app/main.py
```diff
@@ -2,24 +2,83 @@
...
+    def can_access(self, visitor: Visitor) -> bool:
+        if not isinstance(visitor, Visitor):
+            raise TypeError("Visitor must be an instance of Visitor class.")
+        return all([
+            self.limitation_validator.age.min_amount
+            <= visitor.age <= self.limitation_validator.age.max_amount,
+            self.limitation_validator.height.min_amount
+            <= visitor.height <= self.limitation_validator.height.max_amount,
+            self.limitation_validator.weight.min_amount
+            <= visitor.weight <= self.limitation_validator.weight.max_amount
```
vsmutok: use `self.limitation_class(...)` instead

app/main.py
```diff
@@ -2,24 +2,83 @@
...
+    ) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__(
```
vsmutok: In ChildrenSlideLimitationValidator you can remove` __init__()`

app/main.py
```diff
@@ -2,24 +2,83 @@
...
+            age=IntegerRange(4, 14),
+            weight=IntegerRange(20, 50),
+            height=IntegerRange(80, 120)
+        )
```
```suggestion
    def __init__(self) -> None:
        age = IntegerRange(4, 14)
        height = IntegerRange(80, 120)
        weight = IntegerRange(20, 50)
```

vsmutok (changes requested):

app/main.py
rzaliznyi: не проходять тести коли це роблю

app/main.py
rzaliznyi: можна трохи ширше пояснити

app/main.py
```diff
@@ -2,24 +2,88 @@
...
+            height=IntegerRange(120, 220)
+        )
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Dimosphen1: Indicate in the type annotation you expect to obtain a subclass of a `SlideLimitationValidator`, not an instance of this class
```suggestion
            limitation_class: type[SlideLimitationValidator]
```

app/main.py
```diff
@@ -2,24 +2,88 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class()
```
Dimosphen1: Do not perform initialization of the `limitation_class` here, perform it in the `can_access` in the try except block where you will catch errors raised in the `__set__` method and return False in case they are cathed and True if not

app/main.py
```diff
@@ -2,24 +2,88 @@
...
+        if not isinstance(visitor, Visitor):
+            raise TypeError("Visitor must be an instance of Visitor class.")
+        return all([
+            self.limitation_class.age.min_amount
+            <= visitor.age <= self.limitation_class.age.max_amount,
+            self.limitation_class.height.min_amount
+            <= visitor.height <= self.limitation_class.height.max_amount,
+            self.limitation_class.weight.min_amount
+            <= visitor.weight <= self.limitation_class.weight.max_amount
+        ])
```
Dimosphen1: Do not perform such comparison, utilize `limitation_class` for check

app/main.py
```diff
@@ -2,24 +2,88 @@
...
+        self.weight = weight
+        self.height = height
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        self.age = IntegerRange(4, 14)
+        self.height = IntegerRange(80, 120)
+        self.weight = IntegerRange(20, 50)
```
Dimosphen1: You could define `age`, `height`, and `weight` as class attributes for `ChildrenSlideLimitationValidator` and `AdultSlideLimitationValidator`

Dimosphen1 (changes requested):

app/main.py
viktoria-rybenchuk: The validation should be provided when you create the limitations class and pass the visitor parameters. Then, the descriptor will raise an error when validating attributes in the `set `method. Hint: use a `try-except` construction.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -2,24 +2,78 @@
...
+        self.public_name = name
+        self.private_name = "_" + name
+
+    def __get__(self, obj: object, objtype: None) -> int:
+        value = getattr(obj, self.private_name)
+        return value
+
+    def __set__(self, obj: object, value: int) -> None:
+        if not isinstance(value, int):
+            raise ValueError("Value must be integer.")
```
Dimosphen1, line 19: In such cases, it would be preferable to raise `TypeError` since type checking is performed. Take this into account in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #482, opened 2024-03-28
Solution

VasylRybalka:


app/main.py
```diff
@@ -1,25 +1,67 @@
...
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: Any, owner: Type) -> Tuple[int, int]:
+        if instance is None:
+            return self.min_amount, self.max_amount
+        return self.min_amount, self.max_amount
```
EdAlekseiev: It must return stored valu from `instance` instead of `self.min_amount, self.max_amount`

app/main.py
```diff
@@ -1,25 +1,67 @@
...
+
+    def __set__(self, instance: Any, value: int) -> None:
+        raise AttributeError("Cannot set value")
```
EdAlekseiev: Why is this method empty?

app/main.py
```diff
@@ -1,25 +1,67 @@
...
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    @abstractmethod
+    def validate(self, visitor: "Visitor") -> bool:
+        pass
```
EdAlekseiev: You should validate data inside `__set__` method of IntegerRange

EdAlekseiev (changes requested):

PR #484, opened 2024-04-02
solved

serhii-chaban:


app/main.py
```diff
@@ -1,25 +1,80 @@
...
+    def __get__(
+            self,
+            obj: SlideLimitationValidator,
+            objtype: Any = None
+    ) -> int:
+        return getattr(obj, self.protected_name)
+
+    def __set__(self, obj: SlideLimitationValidator, value: Any) -> None:
+        if not isinstance(value, int):
+            raise TypeError("Type Error")
```
viktoria-rybenchuk: you should use descriptive error message

app/main.py
```diff
@@ -1,25 +1,80 @@
...
+        if self.min_amount <= value <= self.max_amount:
+            setattr(obj, self.protected_name, value)
+            return
+        raise ValueError("Value Error")
```
viktoria-rybenchuk: you can check not the condition met and then raise `ValueError`, otherwise - set attribute

app/main.py
```diff
@@ -1,25 +1,80 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(
+                visitor.age, visitor.weight, visitor.height
+            )
+            return True
+        except ValueError or TypeError:
```
viktoria-rybenchuk: avoid using `or ` operator

viktoria-rybenchuk (changes requested):

EdAlekseiev (approved):
Good job

PR #491, opened 2024-04-09
Solution!

scandriy:


app/main.py
```diff
@@ -2,24 +2,64 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator
```
ArsenPidhoretskyi: It accepts type of SlideLimitationValidator, not instance

ArsenPidhoretskyi (changes requested):

app/main.py
scandriy: Can't get it. Please explain

```suggestion
                 limitation_class: Type[SlideLimitationValidator]
```

EdAlekseiev (changes requested):

app/main.py
scandriy: Ok, thank you!

EdAlekseiev (approved):

PR #498, opened 2024-04-19
Solution

VladimirDolhyi:


app/main.py
```diff
@@ -2,24 +2,68 @@
...
+
+    def __set_name__(self, name: str) -> None:
+        self.name = name
+
+    def __get__(self) -> tuple:
+        return self.min_amount, self.max_amount
+
+    def __set__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
```
ArsenPidhoretskyi: No no.... Please read again about descriptors
Here should be validation

app/main.py
```diff
@@ -2,24 +2,68 @@
...
+                and self.we.min_amount <= visitor.weight <= self.we.max_amount
+                and self.he.min_amount <= visitor.height <= self.he.max_amount)
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        age = IntegerRange(4, 14)
+        weight = IntegerRange(20, 50)
+        height = IntegerRange(80, 120)
```
ArsenPidhoretskyi: This should be class attributes

app/main.py
```diff
@@ -2,24 +2,68 @@
...
+        super().__init__(age, weight, height)
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        age = IntegerRange(14, 60)
+        weight = IntegerRange(50, 120)
+        height = IntegerRange(120, 220)
```
ArsenPidhoretskyi: This should be class attributes

ArsenPidhoretskyi (changes requested):

app/main.py
VladimirDolhyi: Ok.

app/main.py
VladimirDolhyi: Ok.

app/main.py
VladimirDolhyi: Ok.

app/main.py
```diff
@@ -2,24 +2,84 @@
...
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
+
+    def __get__(self,
+                instance: "SlideLimitationValidator",
+                owner: type
+                ) -> tuple:
```
Dimosphen1: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Apply this change here and in the other places as well

app/main.py
```diff
@@ -2,24 +2,84 @@
...
+        return getattr(instance, self.name)
+
+    def __set__(self,
+                instance: "SlideLimitationValidator",
```
Dimosphen1: Instead of `SlideLimitationValidator` indicate `Any` in the type annotation here and in the `__get__` method to not limit `IntegerRange` applicability

app/main.py
```diff
@@ -2,24 +2,84 @@
...
+                value: int
+                ) -> None | bool:
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError
+        return True
```
Dimosphen1: Use `settatr` here, do not `return True`. Change the corresponding type annotation

app/main.py
```diff
@@ -2,24 +2,84 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
```
Dimosphen1: Utilize protected attribute in the `IntegerRange` class

app/main.py
```diff
@@ -2,24 +2,84 @@
...
+    age = IntegerRange(4, 14)
+    weight = IntegerRange(20, 50)
+    height = IntegerRange(80, 120)
+
+    def __init__(self,
+                 age: IntegerRange,
+                 weight: IntegerRange,
+                 height: IntegerRange
+                 ) -> None:
+        super().__init__(age, weight, height)
```
Dimosphen1: Delete this `__init__` here and in the `AdultSlideLimitationValidator` class as well, you could rely on the parent class implementation.

app/main.py
```diff
@@ -2,24 +2,84 @@
...
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator
```
Dimosphen1: Indicate that you expect to obtain a subclass of a `SlideLimitationValidator` for `limitation_class`, not an instance of the `SlideLimitationValidator` class
```suggestion
                 limitation_class: type[SlideLimitationValidator]
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,25 +1,95 @@
...
+            self,
+            owner: type,
+            name: str
+    ) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(
+            self,
+            instance: Any,
+            owner: type
```
Dimosphen1: It would be better to indicate that `owner` could be `Any` as well

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

app/main.py
VladimirDolhyi: Ok. I have changed it too.

PR #504, opened 2024-05-02
Solution

VasylRybalka:


app/main.py
```diff
@@ -1,25 +1,83 @@
...
+        self.age = Range(4, 14)
+        self.height = Range(80, 120)
+        self.weight = Range(20, 50)
+
+    def validate(self, visitor: Visitor) -> bool:
+        return (
+            self.age.min <= visitor.age <= self.age.max
+            and self.height.min <= visitor.height <= self.height.max
+            and self.weight.min <= visitor.weight <= self.weight.max
+        )
```
EdAlekseiev, line 57: This is redunat method. You should use descriptors for validations these fields. Example https://www.datacamp.com/tutorial/python-descriptors

app/main.py
```diff
@@ -1,25 +1,83 @@
...
+        self.age = Range(14, 60)
+        self.height = Range(120, 220)
+        self.weight = Range(50, 120)
+
+    def validate(self, visitor: Visitor) -> bool:
+        return (
+            self.age.min <= visitor.age <= self.age.max
+            and self.height.min <= visitor.height <= self.height.max
+            and self.weight.min <= visitor.weight <= self.weight.max
+        )
```
EdAlekseiev, line 72: The same comment as above

EdAlekseiev (changes requested):

PR #513, opened 2024-05-23
Solution

b4oody:


app/main.py
```diff
@@ -1,25 +1,63 @@
...
+    height = IntegerRange(min_amount=0, max_amount=0)
+    weight = IntegerRange(min_amount=0, max_amount=0)
+
+    def validate(self, visitor: Visitor) -> bool:
+        valid_age = (
+            self.age.min_amount <= visitor.age <= self.age.max_amount)
+        valid_height = (
+            self.height.min_amount <= visitor.height <= self.height.max_amount)
+        valid_weight = (
+            self.weight.min_amount <= visitor.weight <= self.weight.max_amount)
```
viktoria-rybenchuk: You should perform this validation in the set descriptor method. Each attribute should be checked against its corresponding IntegerRange descriptor, ensuring that the visitor's attributes fall within the specified limits.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,25 +1,77 @@
...
+    def __set__(
+            self, obj: object,
+            value: int
+    ) -> None:
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError(f"Value for "
+                             f"{self.name} must be between "
+                             f"{self.min_amount} and "
+                             f"{self.max_amount}")
+        obj.__dict__[self.name] = value
```
viktoria-rybenchuk: use `setattr` method

app/main.py
```diff
@@ -1,25 +1,77 @@
...
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, obj: object, objtype: type = None) -> int:
+        return obj.__dict__.get(self.name, None)
```
viktoria-rybenchuk: using `getattr` is useful for fetching the value of an attribute when its name is stored in another variable. It also provides a default value if the attribute does not exist, which can help avoid potential AttributeError exceptions.

app/main.py
```diff
@@ -1,25 +1,77 @@
...
+        pass
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def validate(self) -> bool:
+        valid_age = 4 <= self.visitor.age <= 14
+        valid_height = 80 <= self.visitor.height <= 120
+        valid_weight = 20 <= self.visitor.weight <= 50
+        return valid_age and valid_height and valid_weight
```
viktoria-rybenchuk: You can validate values through a descriptor.

app/main.py
```diff
@@ -1,25 +1,77 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def validate(self) -> bool:
+        valid_age = 14 <= self.visitor.age <= 60
+        valid_height = 120 <= self.visitor.height <= 220
+        valid_weight = 50 <= self.visitor.weight <= 120
+        return valid_age and valid_height and valid_weight
```

viktoria-rybenchuk (changes requested):

app/main.py
vasylhnatiuk: fix it

app/main.py
vasylhnatiuk: fix it

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
.

EdAlekseiev (approved):

PR #514, opened 2024-05-24
solution

AndrewwSSS:


app/main.py
```diff
@@ -1,25 +1,91 @@
...
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+        self._name: str = ""
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self._name = name
+
+    def __get__(self, instance: object, owner: type) -> int:
+        return instance.__dict__.get(self._name)
```
DevSpaciX: You shoud use here getattr()

app/main.py
```diff
@@ -1,25 +1,91 @@
...
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError(f"{self._name.capitalize()} "
+                             f"must be between {self.min_amount} and "
+                             f"{self.max_amount}.")
+        instance.__dict__[self._name] = value
```
DevSpaciX: It is better to use setattr()

app/main.py
```diff
@@ -1,25 +1,91 @@
...
+            return True
+        except ValueError:
+            return False
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: type[SlideLimitationValidator]) -> None:
```
DevSpaciX: You can write it more simple - > `limitation_class: SlideLimitationValidator` is not it ?

app/main.py
```diff
@@ -1,25 +1,91 @@
...
+        pass
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    weight = IntegerRange(20, 50)
+    height = IntegerRange(80, 120)
+
+    def __init__(self) -> None:
```
DevSpaciX: You can expect here` age, weight, height` And then calling super just add the params , instead of hardcoding them 
`super().__init__(age, weight, height)`

app/main.py
```diff
@@ -1,25 +1,91 @@
...
+    age: IntegerRange
+    weight: IntegerRange
+    height: IntegerRange
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+    @abstractmethod
```
DevSpaciX: Is this necessery ? If not delete it please 
image

DevSpaciX (changes requested):

app/main.py
AndrewwSSS: not working without it

app/main.py
AndrewwSSS: not work without this

DevSpaciX (approved):
Great now !

PR #515, opened 2024-05-24
Solution

SuskyiVolodymyr:


app/main.py
```diff
@@ -1,25 +1,96 @@
+from __future__ import annotations
 from abc import ABC
+from typing import Any, Callable
```
Oleg-s-s, line 3: In latest versions of python you can use this without imports, simply write in lowercase `Any` -> `any`

Oleg-s-s (approved):
👍

PR #520, opened 2024-05-29
add solution

jpdreamthug:


app/main.py
```diff
@@ -1,25 +1,89 @@
...
+                "Age should be between 4 and 14 for children."
+            )
+        if not (80 <= self.height <= 120):
+            raise ValueError(
+                "Height should be between 80 and 120 for children."
+            )
+        if not (20 <= self.weight <= 50):
+            raise ValueError(
+                "Weight should be between 20 and 50 for children."
+            )
```
DevSpaciX: Your validation should be in your `__set__ `method ) You just will need to call instances of your `IntegerRange`  class

DevSpaciX (changes requested):

Oleg-s-s (approved):
All good, but personally I dont like the way that you put everything in 1 module. There are a lot of different kinds of logic that can and SHOULD be separated within irl projects

PR #524, opened 2024-06-05
solution

Alex93Troy:


app/main.py
```diff
@@ -2,24 +2,70 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __get__(self, instance: object, owner: object) -> object:
+        return instance.__dict__[self.name]
```
vsmutok: use `getattr` here

app/main.py
```diff
@@ -2,24 +2,70 @@
...
+
+    def __set__(self, instance: object, value: str) -> None:
+        instance.__dict__[self.name] = value
```
vsmutok: use `setattr` here

vsmutok (changes requested):

app/main.py
Alex93Troy: okay, thanks i add

vsmutok (approved):
Good Job!

PR #531, opened 2024-06-09
created classes related to aquapark

Timur5050:


app/main.py
```diff
@@ -2,24 +2,68 @@
...
+        self.protected_field = "_" + name
+
+    def __get__(self, instance: any, owner: any) -> int:
+        return getattr(instance, self.protected_field)
+
+    def __set__(self, instance: any, value: any) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_field, value)
+        else:
+            setattr(instance, self.protected_field, None)
```
viktoria-rybenchuk: Raising a `ValueError` makes it explicit that the value provided is invalid.

app/main.py
```diff
@@ -2,24 +2,68 @@
...
+            age=visitor.age,
+            height=visitor.height,
+            weight=visitor.weight
+        )
+        if (visitor_result.age
+                and visitor_result.weight
+                and visitor_result.height):
+            return True
+
+        return False
```
viktoria-rybenchuk: You can catch the ValueError here with the help of construction `try-except`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -2,24 +2,66 @@
...
+    def __set_name__(self, owner: any, name: any) -> None:
+        self.protected_field = "_" + name
+
+    def __get__(self, instance: any, owner: any) -> int:
+        return getattr(instance, self.protected_field)
+
+    def __set__(self, instance: any, value: any) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_field, value)
+        else:
```
viktoria-rybenchuk: you can check if the value not in the range and then raise ValueError, otherwise set attribute then you avoid using else.
Add descriptive error message to the `ValueError`

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Great job!

PR #540, opened 2024-07-02
Add solution

MRozdoba:


app/main.py
```diff
@@ -1,25 +1,72 @@
...
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError("Value should be an integer")
+        if not self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, "access denied")
```
viktoria-rybenchuk: If there is an incorrect value, you should raise an error without setting the attribute

app/main.py
```diff
@@ -1,25 +1,72 @@
...
+            )
+        else:
+            permission = ChildrenSlideLimitationValidator(
+                visitor.age, visitor.weight, visitor.height
+            )
+        if (
+                permission.weight == "access denied"
+                or permission.height == "access denied"
+                or permission.age == "access denied"
+        ):
```
viktoria-rybenchuk: you can create permission by passing attributes to the `self.limitation_class`
and write this logic by using [`try-except`](https://www.w3schools.com/python/python_try_except.asp) construction, when you have caught the error then return False, else True

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,25 +1,62 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
```
Dimosphen1: You should indicate in the type annotation that you expect to obtain a subclass of `SlideLimitationValidator` class here
```suggestion
    def __init__(self, name: str, limitation_class: type[SlideLimitationValidator]) -> None:
```

app/main.py
```diff
@@ -1,25 +1,62 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
+            return False
+        else:
```
Dimosphen1: You could omit using `else` here

app/main.py
```diff
@@ -1,25 +1,62 @@
...
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
```
Dimosphen1: Since you raise `TypeError` as well, catch it in this `except` statement along with `ValueError`

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+from typing import Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
```
pavlopro: Please set `object ` as type annotation instead of Any

app/main.py
pavlopro:
```
      print("Access denied or incorrect data provided")
      return False
```

pavlopro (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #549, opened 2024-07-13
Solution

Prorochenkko:


YukihiroSM (approved):
Good job! But try to think on how can you escape duplicating validate method in ChildrenSlideLimitationValidator and AdultSlideLimitationValidator

PR #550, opened 2024-07-14
aqua-park implementation

MasakDirt:


app/main.py
```diff
@@ -1,25 +1,75 @@
...
+                 limitation_class: SlideLimitationValidator) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
+            return False
+        return True
```
oliasymak24, line 80: Do you need `return True` after the exception block?

app/main.py
```diff
@@ -1,25 +1,75 @@
...
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
```
oliasymak24: To avoid incorrect assignment of arguments you can use keywords arguments.

app/main.py
```diff
@@ -1,25 +1,75 @@
...
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
```
oliasymak24: Pay attention to absctract __init__  method. Is it mandatory to define each attribute?

oliasymak24 (changes requested):
In general your code looks code but pay attention to some comments :)

app/main.py
MasakDirt: I think yes, if I return anything there would return None

app/main.py
MasakDirt: I think no)

MasakDirt:
> In general your code looks code but pay attention to some comments :)

Well, thanks but why on a platform you reject this task?)

app/main.py
```diff
@@ -1,25 +1,76 @@
...
+        super().__init__(age, weight, height)
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator) -> None:
```
MNDonut: Read this message please
https://mate.academy/chats/244801?selectedMessageId=3129519

MNDonut (changes requested):

app/main.py
MasakDirt: I wrote implementation before I learn how to write params in methods/functions)

app/main.py
oliasymak24: You can put it directly to `try` block

oliasymak24 (approved):

PR #551, opened 2024-07-14
Solution py-aquapark

vldyu:


app/main.py
```diff
@@ -2,24 +2,82 @@
...
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> int:
+        value = getattr(instance, self.protected_name)
+        return value
+
+    def __set__(self, instance: object, value: type) -> None:
```
pavlopro: So give a hint to developers that value should be an int value

app/main.py
```diff
@@ -2,24 +2,82 @@
...
+        self.height = height
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    weight = IntegerRange(20, 50)
+    height = IntegerRange(80, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
```
pavlopro: Remove those `inits` definition in child classes

app/main.py
```diff
@@ -2,24 +2,82 @@
...
+            name: str,
+            limitation_class: SlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
```
pavlopro: How about `TypeError`?

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #557, opened 2024-07-23
Solution

DSahalatyi:


app/main.py
```diff
@@ -2,24 +2,82 @@
...
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(
+            self,
+            owner: "SlideLimitationValidator",
```
MNDonut: don't use quotes, you can import annotations from future and it'll not show you errors

app/main.py
```diff
@@ -2,24 +2,82 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
MNDonut: no need to call init if you just need to call super.init, it's called automatically

MNDonut (changes requested):

oliasymak24 (approved):
Good job!

PR #559, opened 2024-07-28
Honestly, I don't understand how to make this task some other way, bu…

BastovOleksandr:
Honestly, I don't understand how to make this task some other way, but I'm sure the current solution is wrong. I need to "chain" these validators and descriptors differently.
Moreover, I have used the IntegerRange descriptor in two child classes but not in the parent one, which could be more laconic and without code repeating.
And validate() methods are the same in both child classes....

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -19,8 +19,7 @@ class IntegerRange:
         return getattr(instance, self.protected_name)
 
     def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
-        if not (isinstance(value, int)
-                and self.min_amount <= value <= self.max_amount):
+        if not self.min_amount <= value <= self.max_amount:
             value = None
         setattr(instance, self.protected_name, value)
 
@@ -34,15 +33,17 @@ class Visitor:
 
 
 class SlideLimitationValidator(ABC):
-    def __init__(self, age: int, height: int, weight: int) -> None:
-        self.age = age
-        self.height = height
-        self.weight = weight
-
-    def validate(self) -> bool:
-        if self.age and self.weight and self.height:
-            return True
-        return False
+    def __init__(
+            self,
+            age: int = None,
+            height: int = None,
+            weight: int = None
+    ) -> None:
+        if all(isinstance(value, int) for value in (age, height, weight)):
+            self.age = age
+            self.height = height
+            self.weight = weight
+            self.approved = all((self.age, self.weight, self.height))
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,25 +1,75 @@
...
+        self,
+        instance: SlideLimitationValidator,
+        owner: SlideLimitationValidator
+    ) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
+        if not (isinstance(value, int)
+                and self.min_amount <= value <= self.max_amount):
+            value = None
```
MNDonut: Read this message please
https://mate.academy/chats/244801?selectedMessageId=3129519

app/main.py
```diff
@@ -1,25 +1,75 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+        self,
+        name: str,
+        limitation_class: type
```
MNDonut: instead of type you can use SlideLimitationValidator because all your validators inherit from it

app/main.py
```diff
@@ -1,25 +1,75 @@
...
-    pass
+    def __init__(self, age: int, height: int, weight: int) -> None:
+        self.age = age
+        self.height = height
+        self.weight = weight
+
+    def validate(self) -> bool:
+        if self.age and self.weight and self.height:
+            return True
+        return False
```
MNDonut: this validate should check if visitor's age, height and weight is correct according to age height weight of this validator

app/main.py
```diff
@@ -1,25 +1,75 @@
...
+    def __get__(
+        self,
+        instance: SlideLimitationValidator,
+        owner: SlideLimitationValidator
+    ) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
+        if not (isinstance(value, int)
+                and self.min_amount <= value <= self.max_amount):
```
MNDonut: It should just set value. as it's integer range class, values should be integers, thus check it

MNDonut (changes requested):

app/main.py
```diff
@@ -1,25 +1,75 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(
+        self,
+        name: str,
+        limitation_class: SlideLimitationValidator()
```
oliasymak24: Do you need brackets here?

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #560, opened 2024-07-31
implement solution for aqua-park task

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1,25 +1,62 @@
...
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected = "_" + name
+
+    def __get__(self, instance: SlideLimitationValidator, owner: type) -> int:
+        return getattr(instance, self.protected)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
+        if not isinstance(value, int):
```
Oleksl888, line 25: It is a better practice to raise exception here with a descriptive message that tells why the validation didn't pass. You haven't studied it yet so it is fine to keep code as is. Here's an example how could have this been done:
```
            raise ValueError(f"Value must be between {self.min_amount} and {self.max_amount}")

File: app/main.py [hunk 2]
```diff
@@ -1,25 +1,62 @@
...
+            return
+        if not (self.min_amount <= value <= self.max_amount):
+            return
+        setattr(instance, self.protected, value)
 
 
 class Visitor:
-   pass
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
```
Oleksl888: Too many params on one line, even though it fits within 79 symbol limit it is still better to put each on one line

app/main.py
```diff
@@ -1,25 +1,62 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
```
Oleksl888: You can specify here that SlideLimitationValidator is exepcted here - but mention it not as an instance but as class itself

app/main.py
```diff
@@ -1,25 +1,62 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        limitations_passed = self.limitation_class(
```
Oleksl888: Going back to exceptions, if you have implemented validator with exception you could have try and catch it here which would make your code more clear:
```
        try:
            self.limitation_class(
                age=visitor.age,
                weight=visitor.weight,
                height=visitor.height
            )
        except ValueError:
            return False
        return True

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #562, opened 2024-08-02
Solution

Narberal90:


app/main.py
```diff
@@ -1,25 +1,78 @@
...
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: Any, value: int) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, value)
+        else:
```
oliasymak24: you don't need this else, raise an error if not  self.min_amount <= value <= self.max_amount will be better

app/main.py
```diff
@@ -1,25 +1,78 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
+        return getattr(instance, self.protected_name)
```
oliasymak24, line 14: I thing this method will return a string

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+        self.height = height
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
```
oliasymak24: You don't need here an __init__, check in other places too

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class Slide:
-    pass
+    def __init__(
+        self, name: str,
+        limitation_class: type(SlideLimitationValidator)
```
oliasymak24: just SlideLimitationValidator will be enough, without type

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #563, opened 2024-08-04
Solution

zazmarga:


app/main.py
```diff
@@ -2,24 +2,72 @@
...
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator
+                 ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        class_visit = None
+        valid_adult = AdultSlideLimitationValidator(visitor.age,
```
vasylhnatiuk: reuse self.limitation_clas

vasylhnatiuk (changes requested):

app/main.py
zazmarga: I think so

app/main.py
```diff
@@ -2,24 +2,71 @@
...
+
+    def __init__(self, name: str, age: int, weight: int, height: int) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+        self.height = height
+        self.limitation_class_ch = (ChildrenSlideLimitationValidator
+                                    (self.age, self.weight, self.height))
+        self.limitation_class_ad = (AdultSlideLimitationValidator
+                                    (self.age, self.weight, self.height))
```
vsmutok: this attributes redundant here

app/main.py
```diff
@@ -2,24 +2,71 @@
...
+                == len(visitor.limitation_class_ad.__dict__)
+                == 3):
+            return True
+        if len(visitor.limitation_class_ad.__dict__) == 3:
+            class_visit = type(visitor.limitation_class_ad)
+        if len(visitor.limitation_class_ch.__dict__) == 3:
+            class_visit = type(visitor.limitation_class_ch)
+        if self.limitation_class == class_visit:
+            return True
+        return False
```
vsmutok: better use `try ... except` here

app/main.py
```diff
@@ -2,24 +2,71 @@
...
+        self.protected_name = "_" + name
 
-class Visitor:
-   pass
+    def __get__(self, instance: str, owner: str) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: str, value: int) -> None:
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.protected_name, value)
```
vsmutok: raise an error here if the data is not valid

vsmutok (changes requested):

app/main.py
```diff
@@ -2,24 +2,73 @@
...
+                return False
+        if (self.limitation_class.__name__
+                == "ChildrenSlideLimitationValidator"):
+            try:
+                ChildrenSlideLimitationValidator(visitor.age,
+                                                 visitor.weight,
+                                                 visitor.height)
+            except ValueError:
+                return False
+        return True
```
```suggestion
        try:
            self.limitation_class(visitor.age, visitor.weight, visitor.height)
            return True
        except ValueError:
            return False
```

app/main.py
```diff
@@ -2,24 +2,73 @@
...
+        self.age = age
+        self.weight = weight
+        self.height = height
 
 
 class Slide:
-    pass
+    def __init__(self, name: str,
+                 limitation_class: SlideLimitationValidator
+                 ) -> None:
```
```suggestion
    def __init__(
        self, name: str,
        limitation_class: Type[SlideLimitationValidator]
    ) -> None:
```

app/main.py
```diff
@@ -2,24 +2,73 @@
...
+        self.protected_name = "_" + name
 
-class Visitor:
-   pass
+    def __get__(self, instance: str, owner: str) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: str, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError
```
```suggestion
            raise ValueError('some message')
```

vsmutok (changes requested):

app/main.py
zazmarga: Thanks, because before I couldn't understand why Pycharm gave me an error here: "elf.limitation_class(visitor.age, visitor.weight, visitor.height) - not callable"

EdAlekseiev (approved):

PR #564, opened 2024-08-04
Solution Politov

Alexandr-Politov:


app/main.py
```diff
@@ -1,25 +1,70 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: callable) -> None:
```
oliasymak24: limitation_class has  SlideLimitationValidator type

oliasymak24 (changes requested):

app/main.py
Alexandr-Politov: but if I use this type, PyCharm underlines the following code and says, that it's type can be only callable
image

oliasymak24: Seems that there are some problem with Python version and Pycharm version, you can just push it like SlideLimitationValidator, I'll approve it)

oliasymak24 (approved):

PR #565, opened 2024-08-04
Solution

olena4reunited:


app/main.py
```diff
@@ -2,24 +2,69 @@
...
+    def __set_name__(self, owner: type, name: str) -> None:
+        self._protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> int:
+        return getattr(instance, self._protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if (self.min_amount <= value <= self.max_amount):
+            setattr(instance, self._protected_name, value)
+        else:
```
oliasymak24: You don't need this `else` here, better to raise an error it not (self.min_amount <= value <= self.max_amount)

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #571, opened 2024-08-20
'Solution'

ABaidin:


app/main.py
```diff
@@ -1,25 +1,113 @@
...
+            and (
+                self.weight_range.min_amount
+                <= visitor.weight
+                <= self.weight_range.max_amount
+            )
+            and (
+                self.height_range.min_amount
+                <= visitor.height
+                <= self.height_range.max_amount
+            )
```
oliasymak24: No need to impement this method twice, you can implement it in parents class

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,25 +1,90 @@
...
+        height_range = IntegerRange(120, 220)
+        super().__init__(age_range, weight_range, height_range)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
```
Oleksl888: You should annotate this as Type[SlideLimitationValidator] to signify this is a class - not  an instance of a class

app/main.py
```diff
@@ -1,25 +1,90 @@
...
+            self,
+            age: IntegerRange,
+            weight: IntegerRange,
+            height: IntegerRange
+    ) -> None:
+        self.age_range = age
+        self.weight_range = weight
+        self.height_range = height
+
+    def validate(self, visitor: Visitor) -> bool:
```
Oleksl888: this validation happens in your descriptor set method - you don't need it here

app/main.py
```diff
@@ -1,25 +1,90 @@
...
+    def __init__(
+            self,
+            name: str,
+            limitation_class: SlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class()
+
+    def can_access(self, visitor: Visitor) -> bool:
+        return self.limitation_class.validate(visitor)
```
Oleksl888: you don't need to call method validate here  - it would be enough simply to pass the data into limitation class. Also this should be wrapped in try-excpet block and return true or false based on if the exception was raised or not

Oleksl888 (changes requested):

ABaidin:
oh...

Oleksl888 (approved):
You got it ! 🎉

PR #581, opened 2024-08-29
'Solution'

TsapenkoA:


app/main.py
```diff
@@ -1,25 +1,79 @@
...
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def validate(self, value: int) -> None:
+        if value < self.min_amount or value > self.max_amount:
+            raise ValueError(
+                f"Value {value} is out of range"
+                f" ({self.min_amount}, {self.max_amount})"
+            )
```
pavlopro, line 14: The `IntegerRange` class needs to implement the descriptor protocol.

app/main.py
```diff
@@ -1,25 +1,79 @@
...
+        self.name = name
+        self.age = age
+        self.height = height
+        self.weight = weight
 
 
 class SlideLimitationValidator(ABC):
-    pass
+    def __init__(
+            self,
```
pavlopro: `There should be only 4 parameters including self`

app/main.py
```diff
@@ -1,25 +1,79 @@
...
+            max_height: int,
+            min_weight: int,
+            max_weight: int
+    ) -> None:
+        self.age_range = IntegerRange(min_age, max_age)
+        self.height_range = IntegerRange(min_height, max_height)
+        self.weight_range = IntegerRange(min_weight, max_weight)
+
+    @abstractmethod
+    def can_access(self, visitor: Visitor) -> bool:
```
pavlopro: There shouldn't be `can_access` method here

app/main.py
```diff
@@ -1,25 +1,79 @@
...
+        super().__init__(14, 60, 120, 220, 50, 120)
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.age_range.validate(visitor.age)
+            self.height_range.validate(visitor.height)
+            self.weight_range.validate(visitor.weight)
+            return True
+        except ValueError:
+            return False
```
pavlopro: `ChildrenSlideLimitationValidator` and `AdultSlideLimitationValidator` classes should have only class attributes instead

pavlopro (changes requested):

vsmutok (approved):
Looks Good!

PR #585, opened 2024-09-08
Solution

fposlushnyi:


app/main.py
```diff
@@ -2,24 +2,85 @@
...
+    def __set__(self, obj: any, value: int) -> None:
+        self.validate(value)
+        setattr(obj, self.protected_name, value)
+
+    def validate(self, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError("Expected value to be an int")
+
+        if value < self.min_amount or value > self.max_amount:
+            raise ValueError("Value error !!!")
```
Dimosphen1: Make the error message more descriptive rather than duplicating the actual error name

app/main.py
```diff
@@ -2,24 +2,85 @@
...
+    weight = IntegerRange(min_amount=50, max_amount=120)
 
 
 class Slide:
-    pass
+    def __init__(
+        self,
+        name: str,
+        limitation_class: ChildrenSlideLimitationValidator
+            | AdultSlideLimitationValidator
```
Dimosphen1: Indicate that you expect to obtain a subclass of `SlideLimitationValidator` class rather than instances of `ChildrenSlideLimitationValidator` or `AdultSlideLimitationValidator` classes
```suggestion
        limitation_class: type[SlideLimitationValidator]
```

app/main.py
```diff
@@ -2,24 +2,85 @@
...
+                    visitor.age,
+                    visitor.height,
+                    visitor.weight
+                )
+            except ValueError:
+                has_access = False
+            else:
+                has_access = True
+
+        return has_access
```
Dimosphen1: Try to initialize `self.limitation_class` with required values and in case of error return `False` and `True` otherwise. Remove the redundant logic

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+        try:
+            self.limitation_class(
+                visitor.age,
+                visitor.height,
+                visitor.weight
+            )
+        except ValueError:
+            return False
+        else:
+            return True
```
```suggestion
            return False
        return True
```
`else` redundant in this case

app/main.py
```diff
@@ -2,24 +2,75 @@
...
+
+    def __set_name__(self, owner: any, name: any) -> None:
+        self.protected_name = "_" + name
+
+    def __get__(self, obj: any, objtype: any = None) -> None:
+        return getattr(obj, self.protected_name)
+
+    def __set__(self, obj: any, value: int) -> None:
+        self.validate(value)
+        setattr(obj, self.protected_name, value)
```
vsmutok: The annotation for the methods is incorrect. Please correct it.

vsmutok (changes requested):

app/main.py
```diff
@@ -2,24 +2,74 @@
...
+        if value < self.min_amount or value > self.max_amount:
+            raise ValueError(
+                f"Expected {value!r} to be no more than {self.min_amount}"
+                f"and less than {self.max_amount}."
+            )
 
 
 class Visitor:
-   pass
+    def __init__(self, name: int, age: int, height: int, weight: int) -> None:
```
Dimosphen1, line 31: In case there are multiple arguments, you could place each on a new line to improve the readability

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #593, opened 2024-09-22
Solution_1

apmaz:


app/main.py
```diff
@@ -1,25 +1,80 @@
...
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Visitor, name: str) -> None:
+        self.name = "_" + name
+
+    def __get__(
+            self,
+            instance: SlideLimitationValidator,
+            owner: Visitor
+    ) -> Visitor:
```
Dimosphen1: Please, use more generic types in the type annotation to use `IntegerRange` class for classes other than indicated here as well (e.g. use `Any`)

app/main.py
```diff
@@ -1,25 +1,80 @@
...
+        return getattr(instance, self.name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
```
Dimosphen1, line 21: Raise exception in the `IntegerRange` class to track it presence during the class initialization

app/main.py
```diff
@@ -1,25 +1,80 @@
...
+            visitor.height
+        )
+
+        if False in [
+            limitation_class_instance.age,
+            limitation_class_instance.weight,
+            limitation_class_instance.height
+        ]:
+            return False
+        return True
```
Dimosphen1: Use try except block here to catch an exception raised during the `IntegerRange` class initialization and return the corresponding boolean value based on the exception presence

app/main.py
```diff
@@ -1,25 +1,80 @@
...
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: type
```
Dimosphen1: Indicate that you expect to obtain subclasses of the `SlideLimitationValidator` class for `limitation_class` in the type annotation, please

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+    def __get__(
+            self,
+            instance: Any,
+            owner: Visitor
+    ) -> Visitor:
+        return getattr(instance, self.name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
+        if value not in range(self.min_amount, self.max_amount + 1):
+            raise ValueError
```
```suggestion
            raise ValueError("some message")
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+        self.name = "_" + name
+
+    def __get__(
+            self,
+            instance: Any,
+            owner: Visitor
+    ) -> Visitor:
+        return getattr(instance, self.name)
+
+    def __set__(self, instance: SlideLimitationValidator, value: int) -> None:
```
Dimosphen1, line 21: You could use `Any` for `instance` and `owner` parameters as well in the type annotation

app/main.py
```diff
@@ -1,25 +1,69 @@
...
+        if value not in range(self.min_amount, self.max_amount + 1):
+            raise ValueError("Sorry the visitor can't use the slide")
```
Dimosphen1, line 23: The error message could be more generic without linking to a specific visitor, e.g.
```suggestion
            raise ValueError(f"{value} is not within range [{self.min_count}, {self.max_count}]")
```

Dimosphen1 (approved):
LGTM, several comments were left for consideration

PR #595, opened 2024-09-26
Solution

KKvladys:


dmytrik (approved):

app/main.py
```diff
@@ -2,24 +2,77 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
```
AnyoneClown: We don't need to create `self.name` here

app/main.py
```diff
@@ -2,24 +2,77 @@
...
+    ) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except ValueError:
+            return False
+        return True
```
AnyoneClown: You have a check for 'ValueError', but there may also be a 'TypeError', which you raise in IntegerRange

AnyoneClown (changes requested):

app/main.py
```diff
@@ -7,15 +7,12 @@ def __init__(self, min_amount: int, max_amount: int) -> None:
...
-        self.name = name
         self.protected_name = "_" + name
 
     def __get__(self, instance: object, owner: type) -> int:
         return getattr(instance, self.protected_name)
 
     def __set__(self, instance: object, value: int) -> None:
-        if not isinstance(value, int):
-            raise TypeError("The value must be an integer..")
         if not (self.min_amount <= value <= self.max_amount):
```
AnyoneClown, line 18: No, that part was correct. You just needed to add a check for `TypeError` in your `try/except`

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #599, opened 2024-10-04
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -1,25 +1,78 @@
...
+        self.height_range = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(
+            self, name: str, limitation_class: SlideLimitationValidator
+    ) -> None:
+        self.name = name
+        self.validator = limitation_class()
```
AnyoneClown: You don't need to call `limitation_class` in your `__init__` method. Read about `Delegation` or `Dependency injection`. You expect an already created class object in init, so the brackets are unnecessary there

```suggestion
        self.validator = limitation_class
```

[Link](https://python-dependency-injector.ets-labs.org/introduction/di_in_python.html)

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+        super().__init__()
+        self.age_range = IntegerRange(4, 14)
+        self.weight_range = IntegerRange(20, 50)
+        self.height_range = IntegerRange(80, 120)
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__()
```
AnyoneClown: You don't need `__init__` and `super().__init__` here, if you are not accepting anything

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+            <= self.weight_range.max_amount,
+            self.height_range.min_amount <= visitor.height
+            <= self.height_range.max_amount
+        ))
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    def __init__(self) -> None:
+        super().__init__()
```
AnyoneClown: The same here

app/main.py
```diff
@@ -1,25 +1,78 @@
...
+
+    def validate_visitor(self, visitor: Visitor) -> bool:
+        return all((
+            self.age_range.min_amount <= visitor.age
+            <= self.age_range.max_amount,
+            self.weight_range.min_amount <= visitor.weight
+            <= self.weight_range.max_amount,
+            self.height_range.min_amount <= visitor.height
+            <= self.height_range.max_amount
+        ))
```
AnyoneClown: >3. SlideLimitationValidator class, inherited from ABC class
Its __init__ method takes age, weight, and height.

Read task properly. You don't need to create `IntegerRange` here, you should only accept it from `__init__`

AnyoneClown: Your logic is okay, but we can simplify it. Inside your `Slide` class we can create `try/except` logic., where we will use our `limitation_class` and catch for `ValueError, TypeError`. If u catched error - `return False`, else `return True`

app/main.py
```diff
@@ -1,25 +1,78 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected_name = f"_{name}"
+
+    def __get__(self, instance: object, owner: type) -> IntegerRange:
+        if instance is None:
```
AnyoneClown:
```suggestion
        if not instance:
```

You are returning `int` type here, not `IntegerRange`.

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #601, opened 2024-10-06
Aqua park solution

94nj111:


app/main.py
```diff
@@ -2,24 +2,69 @@
...
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError(f"{self.public_name.capitalize()} "
+                            f"should be integer")
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError(f"{self.public_name.capitalize()} should be "
+                             f"greater than {self.min_amount} "
+                             f"and less than {self.max_amount}")
```
vladyslav-tmf: This check works correctly, but the message says that the value must be greater than `{self.min_amount}` and less than `{self.max_amount}`, but this isn't quite true, so it also means that the check should look like this:
`if not self.min_amount < value < self.max_amount`.

app/main.py
```diff
@@ -2,24 +2,69 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: object, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: object) -> None:
```
vladyslav-tmf: the `__get__`,` __set__`, and other methods have general object data types. Define the data types more clearly, for example:
`instance: object, owner: type`

vladyslav-tmf (changes requested):

vladyslav-tmf (approved):
Good job!

app/main.py
```diff
@@ -2,24 +2,69 @@
...
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> None:
+        return getattr(instance, self.protected_name)
```
AnyoneClown: Your method should return `int`. Add it in return annotation

app/main.py
```diff
@@ -2,24 +2,69 @@
...
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: type(SlideLimitationValidator)) -> None:
```
AnyoneClown: You don't need to use `type` for annotation. Just use istance of your Class

```suggestion
                 limitation_class: SlideLimitationValidator) -> None:
```

AnyoneClown (changes requested):
Good, but fix code stlye

AnyoneClown (approved):
LGTM!

PR #602, opened 2024-10-06
Solution

SerhiyDema:
attempt 01

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,25 +1,68 @@
+from __future__ import annotations
 from abc import ABC
+from typing import Any, Type
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.private_name = "_" + name
+
+    def __get__(self, instance: object, owner: Any) -> int:
+        return getattr(instance, self.private_name)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,25 +1,68 @@
...
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        self.private_name = "_" + name
+
+    def __get__(self, instance: object, owner: Any) -> int:
+        return getattr(instance, self.private_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not (self.min_amount <= value <= self.max_amount):
+            value = None
```
```suggestion
        if not isinstance(value, int):
            raise TypeError("The value should be integer.")
        if not self.min_amount <= value <= self.max_amount:
            raise ValueError(
                f"The value should not be less than {self.min_amount} "
                f"and greater than {self.max_amount}."
            )
```

app/main.py
```diff
@@ -1,25 +1,68 @@
...
+            return False
+        validate_instance = self.limitation_class()
+
+        validate_instance.age = visitor.age
+        validate_instance.height = visitor.height
+        validate_instance.weight = visitor.weight
+
+        return bool(validate_instance.age
+                    and validate_instance.height
+                    and validate_instance.weight)
```
vsmutok: use `try ... except`  here

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #609, opened 2024-10-11
'Solution'

AndriiDementiev:


app/main.py
```diff
@@ -2,24 +2,60 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
```
Nattalli:
```suggestion
    def __set_name__(self, owner: SlideLimitationValidator, name: str) -> None:
```

app/main.py
```diff
@@ -2,24 +2,60 @@
...
+        self.protected_name = f"_{name}"
+
+    def __get__(self, instance: object, owner: type) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
```
Nattalli: specify `instance` type

app/main.py
```diff
@@ -2,24 +2,60 @@
...
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.protected_name = f"_{name}"
+
+    def __get__(self, instance: object, owner: type) -> int:
```
Nattalli: and here

app/main.py
```diff
@@ -2,24 +2,60 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
```
Nattalli: specify `limitation_class`

Nattalli (changes requested):

app/main.py
```diff
@@ -1,25 +1,63 @@
...
 
 
+class SlideLimitationValidator(ABC):
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        self.age = age
+        self.weight = weight
+        self.height = height
+
+
 class IntegerRange:
```
Arsen-hrynevych, line 11: The IntegerRange descriptor is supposed to work with class attributes in SlideLimitationValidator (age, height, weight), but the current implementation only sets and gets instance attributes.

app/main.py
```diff
@@ -1,25 +1,63 @@
...
 
 class Slide:
-    pass
+    def __init__(self,
+                 name: str,
+                 limitation_class: SlideLimitationValidator) -> None:
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
```
Arsen-hrynevych, line 58: In the can_access method of the Slide class, you instantiate the limitation_class, but you don't check whether the visitor's parameters fit within the range. To properly verify this, you should instantiate the class and then allow the descriptors (age, height, and weight) to validate the visitor's properties by assigning them one by one.

Arsen-hrynevych (changes requested):
Mostly correct.

PR #646, opened 2024-12-07
Solution

julia4406:


app/main.py
```diff
@@ -2,24 +2,69 @@
...
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> None:
+        getattr(instance, self.protected_name)
```
L1nk3rrr: looks like you need return here

L1nk3rrr (changes requested):
A minor point I have found.
Also you can remove your docstring as they repeat your class names, that are very clear what that do, docstrings don't give us more information in this situation, so they are no needed

app/main.py
```diff
@@ -2,24 +2,69 @@
...
 
 class IntegerRange:
-    pass
+    """ class - descriptor for parameters limitations """
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.public_name = name
```
L1nk3rrr: why do you need this attr?

app/main.py
```diff
@@ -2,24 +2,69 @@
...
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> None:
+        getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError
```
L1nk3rrr: what if value not int?
maybe we can check and raise TypeError too?

L1nk3rrr (approved):
Well done! Great work!

PR #659, opened 2024-12-12
solution

DanilTrukhanov:


app/main.py
```diff
@@ -1,25 +1,71 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            if self.limitation_class(
+                    visitor.age,
+                    visitor.weight,
+                    visitor.height
+            ):
```
julia4406: May be IF-statement doesn't need here if we used try-except for checking fails.

julia4406 (approved):

PR #660, opened 2024-12-12
Aquapark validation done

4ikkamony:


app/main.py
```diff
@@ -1,25 +1,98 @@
...
+        self.max_amount = max_amount
+
+    def __get__(self, instance: object, owner: Type[Any]) -> Any:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError(f"Expected an int, got {type(value)}")
+
+        if value not in range(self.min_amount, self.max_amount + 1):
```
L1nk3rrr: using range can be a heavy op, use `a < b < c` construction

app/main.py
```diff
@@ -1,25 +1,98 @@
 from abc import ABC
+from typing import Type, Any
```
L1nk3rrr: You don't need to import Type, use from __future__ import annotations

L1nk3rrr (changes requested):
Well done! I have left minor points for improvement of your code

L1nk3rrr (approved):
Well done!

PR #663, opened 2024-12-14
Solution

ruslanyemets:


app/main.py
```diff
@@ -2,24 +2,60 @@
...
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: str, name: str) -> None:
+        self.public_name = name
+        self.protected_name = "_" + name
+
+    def __get__(self, instance: "IntegerRange", owner: str) -> int:
```
L1nk3rrr: import `from __future__ import annotations` to use a class name, not a string

app/main.py
```diff
@@ -2,24 +2,60 @@
...
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: "IntegerRange", value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError
+        if value not in range(self.min_amount, self.max_amount + 1):
```
L1nk3rrr: making a range can be too complicated operation, use `<` and `>`, you can use constructions that we used in school, like `a <= b < c`

app/main.py
```diff
@@ -2,24 +2,60 @@
...
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    weight = IntegerRange(50, 120)
+    height = IntegerRange(120, 220)
 
 
 class Slide:
-    pass
+    def __init__(self, name: str, limitation_class: type) -> None:
```
L1nk3rrr: limitation_class is `SlideLimitationValidator` type, not?

app/main.py
L1nk3rrr: also owner is not str type, it can be generalized for `object`

app/main.py
```diff
@@ -2,24 +2,60 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: str, name: str) -> None:
```
L1nk3rrr: owner is not str

L1nk3rrr (changes requested):
Well done! I left some points for improvement for your code!

L1nk3rrr (approved):
Well done!

PR #668, opened 2024-12-19
Solution

taraskhom:


app/main.py
```diff
@@ -2,24 +2,89 @@
...
+    ) -> int:
+        return getattr(instance, self.private_name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError
+        if self.min_amount <= value <= self.max_amount:
+            setattr(instance, self.private_name, value)
+        else:
+            raise ValueError
```
L1nk3rrr: Looks like used redundant else. You can try to use `if not (...)` and raise error, after that if place set attr, so if error not raised you can set it!

L1nk3rrr (changes requested):
Great work! Minor point for improvement was left

L1nk3rrr (approved):
Well done!

PR #691, opened 2025-02-04
solution

Bleit1:


app/main.py
```diff
@@ -2,24 +2,118 @@
...
+                    visitor.age,
+                    visitor.weight,
+                    visitor.height
+                )
+            else:
+                AdultSlideLimitationValidator(
+                    visitor.age,
+                    visitor.weight,
+                    visitor.height
+                )
```
L1nk3rrr, line 116: You don't need `if/else` at all you can simply use 
```
 self.limitation_class(
                    visitor.age,
                    visitor.weight,
                    visitor.height
                )
```

app/main.py
```diff
@@ -2,24 +2,118 @@
...
+    ) -> None:
+        super().__init__(age, weight, height)
 
 
 class Slide:
-    pass
+    def __init__(
+            self,
+            name: str,
+            limitation_class: type
```
L1nk3rrr, line 98: it's not a type, you can put here SlideLimitationValidator, because they all was inherited from it

app/main.py
```diff
@@ -2,24 +2,118 @@
...
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(
+            self,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        super().__init__(age, weight, height)
```
L1nk3rrr, line 77: don't need to rewrite init

app/main.py
```diff
@@ -2,24 +2,118 @@
...
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(
+            self,
+            age: int,
+            weight: int,
+            height: int
+    ) -> None:
+        super().__init__(age, weight, height)
```
L1nk3rrr, line 91: don't need to rewrite init

app/main.py
```diff
@@ -2,24 +2,118 @@
...
+                )
+            else:
+                AdultSlideLimitationValidator(
+                    visitor.age,
+                    visitor.weight,
+                    visitor.height
+                )
+            return True
+        except ValueError:
+            return False
```
L1nk3rrr, line 119: catch Value and Type errors here

app/main.py
```diff
@@ -2,24 +2,118 @@
...
+
+    def __set_name__(
+            self,
+            owner: object,
+            name: str
+    ) -> None :
+        self.protected_name = f"_{name}"
+
+    def validate_amount(self, value: int) -> None:
+        if not type(value) == int:
```
L1nk3rrr, line 21: use `isinstance` instead

L1nk3rrr (changes requested):
Left points for improvement

PR #710, opened 2025-03-03
solution

iSevenpwnz:


Roksolana-K (approved):
Nice job! However, I see that you have one comment on line 15 that I think would be better to remove.

PR #762, opened 2025-06-03
I don't fully understand descriptors and properties.

kravetst:


kravetb (approved):
I don't use it, but it looks good!

PR #771, opened 2025-06-26
Solution

Arch0998:


DrGregoryHouse4676 (approved):

DrGregoryHouse4676 (approved):

app/main.py
```diff
@@ -1,25 +1,92 @@
...
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self,
+                 age: int,
+                 weight: int,
+                 height: int) -> None:
+        super().__init__(age, weight, height)
```
MNDonut: No need to define an empty init. Without explicit implementation python by default makes a super call because when python looks for any method and it isn't implemented in the current class, it'll go to the parent class by MRO

app/main.py
```diff
@@ -1,25 +1,92 @@
...
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self,
+                 age: int,
+                 weight: int,
+                 height: int) -> None:
+        super().__init__(age, weight, height)
```
MNDonut: the same here

MNDonut (changes requested):

MNDonut (approved):

PR #781, opened 2025-07-02
Solution

Plovuezz:


Hetmarum (approved):

Plovuezz:
@mateacademy-ai-mentor you`re wrong: descriptors will still trigger even when self.age = age is done in a base class, as long as the attribute is defined as a descriptor in the actual class (child class). And i dont need to remove those assignments from the base class __init__.

PR #787, opened 2025-07-03
Implemented task

GrishaSeniv:


app/main.py
```diff
@@ -2,24 +2,74 @@
...
 
 
 class ChildrenSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
MNDonut: no need to override methods when you just need to call parent's method. it'll be called automatically by MRO (python doesn't find init here -> it goes to SlideLimitationValidator -> finds init and sets values)

app/main.py
```diff
@@ -2,24 +2,74 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
MNDonut: same here

MNDonut (changes requested):

MNDonut (approved):

PR #790, opened 2025-07-04
Solution

zqpiq:


SerhiiBasok (approved):

zqpiq:
@mateacademy-ai-mentor3 In my README it’s clearly stated that no additional abstract methods are required in this class.

Here’s the exact quote from the task description:

SlideLimitationValidator class, inherited from ABC class
Its __init__ method takes age, weight, and height. No additional abstract methods are required.
This is a base class for Validators, so no validation is needed.

So according to the task requirements, adding an abstract method like validate() isn’t necessary here.

PR #793, opened 2025-07-07
Solution

Kostenikov:


Gleg0 (approved):
Well done!

Kostenikov:
@mateacademy-ai-mentor but checklist says "No additional abstract methods are required. This is a base class for Validators, so no validation is needed."

PR #805, opened 2025-07-19
Solution

reinvvay:


SensibleN00B (approved):
good job!
but you can handle both exceptions in one condition like this:
        except (TypeError, ValueError):
            return False

PR #846, opened 2025-10-03
Solution

toomuchtearz:


app/main.py
```diff
@@ -1,25 +1,83 @@
...
+
+    def __get__(self, instance: object, owner: type) -> int:
+        return getattr(instance, self.protected_name)
+
+    def __set__(self, instance: object, value: Any) -> None:
+        if not isinstance(value, int):
+            raise TypeError
+
+        if not (self.min_amount <= value <= self.max_amount):
+            raise ValueError
```
```suggestion
        if not isinstance(value, int):
            raise TypeError("Some informative message....")

        if not (self.min_amount <= value <= self.max_amount):
            raise ValueError("Some informative message....")
```
It is good practice to add an informative message to the error.
Please do this.

vsmutok (changes requested):

app/main.py
toomuchtearz: fixed that, but readme.md says "Error messages can be skipped"

vsmutok (approved):
Looks Good!

PR #862, opened 2025-10-26
'Solution'

Vorongor:


app/main.py
```diff
@@ -2,24 +2,98 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self,
+                 min_amount: int,
+                 max_amount: int
+                 ) -> None:
```
petrykivd: "Self" should be on new line or ") -> None" on the end of previous line
Also check it on another code fragments

app/main.py
```diff
@@ -2,24 +2,98 @@
...
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self,
+                 age: int,
+                 weight: int,
+                 height: int
+                 ) -> None:
+        super().__init__(age, weight, height)
```
petrykivd: If you don't create an init, we will inherit it from the parent class.
Is it needed here?

app/main.py
```diff
@@ -2,24 +2,98 @@
...
+    age = IntegerRange(4, 14)
+    height = IntegerRange(80, 120)
+    weight = IntegerRange(20, 50)
+
+    def __init__(self,
+                 age: int,
+                 weight: int,
+                 height: int
+                 ) -> None:
+        super().__init__(age, weight, height)
```
petrykivd: If you don't create an init, we will inherit it from the parent class.
Is it needed here?

petrykivd (changes requested):

petrykivd (approved):
GJ! 👍

PR #871, opened 2025-11-04
Soullution

Iventyk:


app/main.py
```diff
@@ -2,24 +2,67 @@
...
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
+
+    def __get__(self, instance: object, owner: type) -> object:
+        if instance is None:
+            return self
+        return instance.__dict__[self.name]
```
DonDmytron: using getattr help you don't use isn't necessary checks

app/main.py
```diff
@@ -2,24 +2,67 @@
...
+
+    def __set__(self, instance: object, value: int | str) -> None:
+        if not isinstance(value, int):
+            raise TypeError
+        if value < self.min_amount or value > self.max_amount:
+            raise ValueError
+        instance.__dict__[self.name] = value
```
DonDmytron: setattr will be better here and you don't need raise TypeError

app/main.py
```diff
@@ -2,24 +2,67 @@
 
 
 class IntegerRange:
-    pass
+    def __init__(self, min_amount: int, max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = name
```
DonDmytron: maybe some protected or private name

DonDmytron (changes requested):

app/main.py
```diff
@@ -2,24 +2,63 @@
...
 
 
 class AdultSlideLimitationValidator(SlideLimitationValidator):
-    pass
+    age = IntegerRange(14, 60)
+    height = IntegerRange(120, 220)
+    weight = IntegerRange(50, 120)
+
+    def __init__(self, age: int, weight: int, height: int) -> None:
+        super().__init__(age, weight, height)
```
petrykivd: Maybe we can avoid init?

app/main.py
```diff
@@ -2,24 +2,63 @@
...
+    def __set_name__(self, owner: type, name: str) -> None:
+        self.name = "_" + name
+
+    def __get__(self, instance: object, owner: type) -> object:
+        return getattr(instance, self.name)
+
+    def __set__(self, instance: object, value: int) -> None:
+        if value < self.min_amount or value > self.max_amount:
+            raise ValueError
+        setattr(instance, self.name, value)
```
petrykivd, line 33: Typehint "object"- is a very unclear concept. 
Let's concretize according to the task.

Please check all type hints in the code.

petrykivd (changes requested):

oliasymak24 (approved):

PR #963, opened 2026-03-05
Solution

Reverad:


app/main.py
```diff
@@ -1,25 +1,80 @@
...
+        self.validate(value)
+        setattr(instance, self.protected_name, value)
+
+    def validate(self, value: int) -> None:
+        if not isinstance(value, int):
+            raise TypeError("Value must be an integer")
+        if not self.min_amount <= value <= self.max_amount:
+            raise ValueError(
+                f"Value must be between"
+                f" {self.min_amount} and {self.max_amount}"
```
pawunder1992, line 28: You dont need to add method validate, all this actions you can do in __set__

pawunder1992 (approved):
you can changed the __set__ method. I did this task in another way and i dont know each is right.First i make the solution as you, but after i thinking about "try exept " in future topics.

PR #982, opened 2026-04-03
'Solution'

artradkevych:


app/main.py
```diff
@@ -1,25 +1,75 @@
...
+        self.name = name
+        self.limitation_class = limitation_class
+
+    def can_access(self, visitor: Visitor) -> bool:
+        try:
+            self.limitation_class(visitor.age, visitor.weight, visitor.height)
+        except TypeError:
+            return False
+        except ValueError:
+            return False
```
pawunder1992, line 74: You can write here only: except (TypeError,ValueError):
            return False

pawunder1992 (approved):

PR #1064, opened 2026-08-14
solved

Aragorn-untold:
Very good


app/main.py
```diff
@@ -1,25 +1,71 @@
 from abc import ABC
+from typing import Any
 
 
 class IntegerRange:
-    pass
+    def __init__(self,
```
DonDmytron, line 6: def __init__(
            self,
            min_amount: int,
            max_amount: int
    ) -> None: 
...

looks better

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+                 min_amount: int,
+                 max_amount: int) -> None:
+        self.min_amount = min_amount
+        self.max_amount = max_amount
+
+    def __set_name__(self,
+                     owner: Any,
+                     name: Any) -> None:
+        self.name = name
```
DonDmytron: Why store the name?

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+        self.private_name = "_" + name
```
DonDmytron: 'protected_name' better naming for this

app/main.py
```diff
@@ -1,25 +1,71 @@
...
+
+    def __get__(self, instance: Any, owner: Any) -> Any:
+        return getattr(instance, self.private_name)
+
+    def __set__(self, instance: Any, value: int) -> None:
+        if type(value) is not int:
```
DonDmytron: use logical 'or' here for one check

DonDmytron (changes requested):

DonDmytron (approved):
good job )

